PR #495, opened 2014-10-28
Basic statistics page (ready for synthesis stats)

jimallman:
This page is [available for review](http://devtree.opentreeoflife.org/about/release-history) on **devtree**. It currently has just synthesis stats. Once we have the second stats file `phylesystem.json` to play with, it will be quick to copy the same behavior in the bottom section.

_**NOTE** that the stats file used here is based on the one @pmidford provided, but I added two bogus profiles from "previous months" to work out the UI._ This file (`ot16:/home/opentree/repo/opentree/webapp/static/stats/synthesis.json`) should of course be replaced with real data as it becomes available.
 
I've built a UI to walk through the history of stats profiles, since this will clearly be a requirement. I also added sparklines (tiny plots for each statistic, built on-the-fly using the Google Charts API) as a nod to showing trends.

This branch adds [moment.js](http://momentjs.com/) for parsing ISO-8601 dates from our stats files. I've added this to the **About > Licenses** page in this branch.


webapp/views/layout.html
```diff
@@ -168,6 +171,7 @@
                       <li><a href="{{= URL('about','materials-and-methods') }}">Materials and methods</a></li>
                       -->
                       <li><a href="{{= URL('about','references') }}">Bibliographic references</a></li>
+                      <li><a href="{{= URL('about', 'statistics') }}">Statistics</a></li>
```
jimallman: Let's remove this for now, so we can review the page on Production.

jar398:
Intermediate push to dev / prod for review purposes.  Will leave issue open.  JAR to deploy, KC and everyone else to review, then we'll decide what to do.

webapp/views/layout.html
jar398: This is mooted by the institution of the 'development' branch
https://docs.google.com/document/d/1k1NJhnE4I6Klw7u9RTpbrBgqfhWkvoR8N0fogQ29rEA/edit
.

So please do *not* remove those from the menu in this branch.

jar398:
Superseded by #536

PR #498, opened 2014-11-05
pairs of stacked barcharts showing the stat about OTUs

mtholder:
feel free to reject this one, and recode it. My D3 is ugly.  
As I was working on it, it seemed nicer to have a pair of stacked bar charts for each synthesis:
   - the left one shows were the taxa are in terms of the curation pipeline (in the synth tree / in nominated studies / in non-nominated studies / in OTT but no studies).
  -  the right bar shows the status of OTUs wrt the synthetic tree (# from studies/ # in synth but no studies / # in OTT but not the synth tree).

Note that the python script makes up bogus values when I could not find the data! It emits a warning that the data is bogus, (which shows up on the web page in red text), but we probably want to get on the same page wrt the info that we can expect to be in the stats before we merge to master.

### DIFF
--- a/webapp/controllers/about.py
+++ b/webapp/controllers/about.py
@@ -105,7 +105,7 @@ def otu_statistics():
                          'Unique OTUs in nominated studies': num_otu_in_nominated_studies,
                          'Date': str(date)}
     # sort by date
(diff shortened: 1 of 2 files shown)


jimallman:
This is being shelved in favor of the area chart (above), in a separate `otu-stats-area` branch:
https://github.com/OpenTreeOfLife/opentree/pull/536

PR #536, opened 2015-01-08
Otu stats (area chart version)

jimallman:


jimallman:
NOTE: I need to remove any nav links to this page before we merge!

jimallman:
See useful mockups and discussion in this earlier pull request! https://github.com/OpenTreeOfLife/opentree/pull/498

kcranston:
Comments on the [dev version](https://devtree.opentreeoflife.org/about/otu-statistics):
* I like the area chart!
* chart needs to be much smaller or adapt to browser width
* needs some descriptive text at the top - purpose of page, defn of OTU, etc. 
* needs definitions for each statistic
* needs a description of that icon beside each stat (the little blue line with the red dot)
* the pulldown says synthesis date, but contains far too many time points to be synthesis dates. Perhaps a monthly statistic instead?

jar398:
I think the layout should put the bar chart in about the center of the first screenful, so you see all of it when you browse to the page. There could be an introductory sentence above it, but more detailed explanation should go below (perhaps requiring scrolling) so it's not in the way. This is assuming "ordinary size" browser window (whatever that is) or adaptation.

Having a synthesis date dropdown, and maybe that whole table, might be redundant with the area chart. Maybe the synthesis dates (currently there are only two) should have a special indicator (maybe a vertical bar) in the area chart?

jimallman:
Area chart with event markers (small clickable markers, plus oversize markers in the highlight box):
screen shot 2015-01-12 at 10 08 09 pm

jimallman:
The [latest Progress page on **devtree**](http://devtree.opentreeoflife.org/about/progress) incorporates numerous changes based on feedback and improved statistics (thanks Peter!).

- New colors, derived from the OpenTree “leaf” logo.

- Forcing all timestamps in data to a simple “date” format, with no time component. This gives us uniform granularity for all stats. Note that we already have some intra-day events (esp. in synthesis releases), so this means we ignore some of them. I've opted to keep and show the **last record for a given date** and discard any from earlier in the day.

- Incorporation of a [manually-edited stats file](https://devtree.opentreeoflife.org/static/statistics/ott.json) describing OT Taxonomy versions. This is currently [under version control](https://github.com/OpenTreeOfLife/opentree/blob/otu-stats-area/webapp/static/statistics/ott.json) as well. Should it be?
	
 OTT version dates are based on file modification dates found on the server. The taxon counts are from smasher, using the [adHocReport method](https://github.com/OpenTreeOfLife/reference-taxonomy/blob/eb1abfba8eb34d2da5285b77e8130016e3494bae/org/opentreeoflife/smasher/Taxonomy.java#L2293-L2312) for taxon “cellular organisms". For example, here's the report from **ott2.9**:
 ```python
>>> ott.adHocReport()
cellular organisms
Taxa: 3144026
Tips: 2896060
> Visible: 2263023, hidden: 633037
> Binomial: 2020468, monomial: 174036, other: 701556
...
```
 I'm using the `Tips` value for `taxon_count` in ott.json, and the `Visible` value for `visible_taxon_count`.
	
* Removed all fake data and creeping totals! What you see is the data we actually have, as best I can interpret it.

* More careful placement of markers for synthesis releases and taxonomy versions. This results in far fewer taxonomy markers, but the synthesis markers are a mess. It appears we’re still missing synthesis data for early releases, and there’s a dense series of synthesis data from mid-January to now. Are these experiments, or automatic operations? Are we gathering stats daily, regardless of whether there’s been a new release?

* For now, I’m assuming that a given synthesis release uses the **most recent prior version** of OT taxonomy. This is almost certainly wrong, and we need a way to explicitly store an OTT version number for each synthesis release. I’ve created [an issue](https://github.com/OpenTreeOfLife/opentree/issues/659) for this.

Note that I’ve postponed two of the changes we discussed:

* Changing the main graph to use log scale. The intent here was to better display low levels of coverage, but I wanted to see if we feel the same now that we can see accurate(?) stats on a linear scale.

* Changing the behavior of the “sticky” highlight (for daily stats), so that clicking anywhere on the graph would toggle it between "locked” (stuck to a particular date) and “unlocked” (freely following the mouse). I tried this, and it was confusing to me, in that clicking on one date would lock the highlight there, and clicking on another would unlock it, which felt like a confusing change of behavior. Clicking on a second date simply jumps to that date (and locks the highlight there), as before. I’ve made the “Unlock highlight” button more prominent when it appears.

Comments welcome!

jar398:
The stats directory isn't currently backed up. I guess it should be.

jimallman:
The latest Progress page is up for review on devtree, using the **latest stats from production**. (Is there any reason to keep harvesting dev stats?) A few new features and changes:

- We're using @pmidford's latest stats, which clearly identify the OTT version used in a given synthesis release.

- I've removed most of the synthesis stats gathered by cron, since this should be triggered manually _only_ when building an "official" release.

- I've cleaned up the official taxonomy releases and placed them on the timeline at their original release date (as best we can determine these). There's some chance of misunderstanding if the viewer assumes each synthesis uses the very latest taxonomy, since there's often a lag. But as @jar398 has pointed out, a taxonomy release is a noteworthy event in and of itself, since the latest taxonomy _is_ used in our TNRS and other taxonomic APIs.

- To minimize confusion about the relationship of taxonomy and synth releases, I've added this information in the detailed release pages for each. In the stats tables, you'll find 'Open Tree Taxonomy version used' and 'Used in synthesis releases' entries that link each type of release to the other(s).

- Realizing that @jar398 has designated each synthesis release with a version number (v1, v2, v3), I'm now using this instead of the release date for synth-release URLs and the UI for stepping between them. We still show these dates on the synth-release page, but the version number is now the main identifier.

- We now handle missing stats (from the days before our stats-harvesting tools) more gracefully, showing 'NOT AVAILABLE' where we don't have historical data. This was necessary to include taxonomy versions and early synthesis releases on the timeline. Please review the area chart for the early months and let me know if it looks misleading; in effect, the area chart reads as if no synthesis has yet taken place.

- To show statistics as early in the timeline as possible, I've added an "extra" synthesis release 'v2a - initial stats' on January 16, 2015. The main 'v2' release is on the proper day, but we had no statistics for the period in between.

- I've also added an extra v1 synthesis release, since I've found two different dates for this. Perhaps @jar398 can clarify which is correct.
 
- @kcranston pointed out some display weirdness in Firefox, and I subsequently discovered even stranger things in IE11. I've made some changes to get consistent results across the major browsers.

- I've [moved `synthesis.json` into git](https://github.com/OpenTreeOfLife/opentree/blob/2264175981d3c6ffc5dc3f9f772f8fff380f793a/webapp/static/statistics/synthesis.json), since it now includes some manual additions (unique version identifers and comments). We'll need to either handle this in the automated tools (and release this file from version control) or perhaps re-commit this file as new stats are added for each synth release.

I believe this is now ready for production, but as always, comments and questions are welcome!

jar398:
Re synthesis releases, the following are the only ones I've kept track of: (from ot10:downloads/)

-rw-r--r-- 1 opentree opentree  9555496855 Jul 23  2014 treemachine.db.tgz.20140723
-rw-r--r-- 1 opentree opentree 16226964620 Sep 14  2014 treemachine.db.tgz.20140914
-rw-r--r-- 1 opentree opentree 11485911070 May 14 19:19 treemachine.db.tgz.20150506

The thing that most stands out for me is the huge discontinuous jump in OTUs in phylesystem from Jan 16 to Feb 1 2015. It cannot be correct.

For the benefit of the record and anyone who wasn't there, a number of changes were suggested on today's hangout, including removing 'total OTUs in synthesis' and relabeling 'unique OTUs in synthesis' to reflect that the number does not include taxonomy-only OTUs (I don't remember the wording we picked, something like 'unique OTUs in synthesis from source trees').

kcranston:
Almost.. ready...
Can I request the text at the top of the chart be replaced with the following:

The Open Tree of Life project includes periodic releases of OTT (Open Tree taxonomy) and the latest synthetic tree, as well as incremental addition of published phylogenetic trees. The timeline below shows the increase in the number of OTUs (operational taxonomic units) in various component. Note the large gap between the number of taxa in OTT and the number of taxa in our datastore (phylesystem) and the synthetic tree! Please help close this gap by adding published trees through our [curator interface].

jimallman:
> Can I request the text at the top of the chart be replaced with the following:

Just to confirm: This is on the main Progress page, right?

kcranston:
Finally merging this! w00t!

PR #538, opened 2015-01-10
Tree-view fixes (for disappearing node/edge highlights)

### LINKED ISSUE #537: Node and edge highlights (tree viewer) sometimes disappear
This happens when rolling over the red "?" in these highlights, or moving from one element to another. It seems more prone to failure if we've scrolled or scaled the tree view.

jimallman:
This fixes some very weird behavior when rolling the mouse over node and
edge highlights. Also cleans up some style and uses nicer cursors over
these highlights. Fix #537.


jimallman:
This is available for review on **devtree** (in `development` branch).

Issue #365, opened 2014-07-11
Show MRCA of tree tips (OTUs) for comparison with focal clade


jar398:
When viewing a study and/or tree, compute the MRCA of all the tips/OTUs and display the nearest taxon name (using appropriate or new treemachine service). This is a simple QC step since if it says 'Eukaryota' or 'cellular organisms' then one of the OTUs is probably wrong.

Along the same lines: for any tree, show the subtended subtree of the synthetic tree for those OTUs.

jimallman:
After [further discussion](http://irclog.perlgeek.de/opentreeoflife/2015-01-20), we've decided that saving these test results in Nexson will lead to confusion for other Nexson consumers. I'll make the change to strip these from the study Nexson before saving in the curation app.

jimallman:
Closing this, since all requests above seem to be satisfied.

PR #556, opened 2015-01-20
Strip MRCA-test results from Nexson before saving.

jimallman:
Addresses #365 and OpenTreeOfLife/phylesystem-1#3


jimallman:
This is live now on **devtree**.

Issue #524, opened 2015-01-06
Explain rooted tree vs. "biologically correct" root (curation tool)


jimallman:
_Email from @rgazis:_ 

Even though I rooted the tree (as in the publication) and selected the ingroup. Under trees it says "tree root: arbitrary (not biologically correct)". The same happens with the other studies.

![image](https://cloud.githubusercontent.com/assets/446375/5636437/15048124-95c0-11e4-8d3a-f5c0ca2175df.png)

jimallman:
To make this more clear, I've made the root-setting clickable in the tree list (this is subtle, since it's not really an error):
![clickable-root-setting](https://cloud.githubusercontent.com/assets/446375/5907491/3ff26920-a56d-11e4-9710-3c5699ba213e.png)

Clicking this brings up the tree-popup to the Properties tab. If the user is just viewing the study, this shows an expanded explanation of arbitrary rooting:

![screen shot 2015-01-26 at 3 05 02 pm]
(https://cloud.githubusercontent.com/assets/446375/5907512/7b32845c-a56d-11e4-8b82-dc0bef20113b.png)

A curator who's editing the tree will see the root-setting widget highlighted as we've done for internal node labeling elsewhere:

![screen shot 2015-01-26 at 3 08 05 pm](https://cloud.githubusercontent.com/assets/446375/5907516/82887db0-a56d-11e4-9246-660386fb0a74.png)

jimallman:
This is available for review on **devtree**. @rgazis, does this address your concerns?

jar398:
I'm not too keen on "arbitrary". In fact the root may not be arbitrary - it could actually be the correct one by coincidence, and it is at least likely to b one that is chosen methodically (if 'incorrectly'), not arbitrarily, by an algorithm. I would say in this situation that the correct (biological) root is unknown (to the app), not recorded (in phylesystem), not available (to us), or not curated (by any of our curators). This leads to something like

Biological root is
  o  confirmed by curator
  o  not confirmed (displayed root could be arbitrary)

jimallman:
> not confirmed (displayed root could be arbitrary)

Thanks, will do. As a bonus, this more accurately reflects the logic of the app.

jimallman:
These changes are live on **devtree** for review.

jimallman:
NOTE that this is causing a problem on devtree. Tree popup disappears immediately from this URL:
https://devtree.opentreeoflife.org/curator/study/edit/694?tab=trees

[...]

PR #566, opened 2015-01-28
Click tree root status to show and explain it.

jimallman:
This is important for "arbitrary" rooting, since curators are
surprised to see this setting after choosing a tree's root.
Addresses #524 (see discussion and screenshots there).


jimallman:
We've held this back from merging due to the sudden appearance of a new bug: On opening [some pages](https://devtree.opentreeoflife.org/curator/study/edit/pg_2823?tab=trees) and clicking to show a tree, its popup would appear but immediately hide again.

This appears to be a conflict with the [smart list filters](https://github.com/OpenTreeOfLife/opentree/pull/571), since both manipulate the browser history and want to override one another. I'll try to make these more compatible, but part of that means that we'll **always** manipulate history when opening a popup, even if it's just to show a warning.

jimallman:
OK, the history collision is fixed and working on **devtree**. (In other words, you should be able to [visit this link](https://devtree.opentreeoflife.org/curator/study/edit/pg_2823?tab=trees) and click the tree to see its popup. If the popup immediately hides, please clear your browser's cache and try again.

See commit message above for details.

Issue #494, opened 2014-10-20
Use web2py caching for references list, initial tree views..?


jimallman:
http://web2py.com/books/default/chapter/29/4#cache

jimallman:
Support for tree-view caching (using web2py's RAM cache) was added to **phylesystem-api** in https://github.com/OpenTreeOfLife/phylesystem-api/commit/64242a5d26cfc64d26a85e0407c4c6291f8f2d9a (and cleaned up in a few following commits). Now I'm modifying the webapp's config file to take advantage of this in its method definitions.

jar398:
documented where...?  i.e. how do you use the cached version, and what does
it call (what URL or whatever) to fill the cache.

jimallman:
> documented where...?

I've started documentation for this feature [in the wiki page 'Caching results from common API calls'](https://github.com/OpenTreeOfLife/opentree/wiki/Caching-results-from-common-API-calls). Comments welcome!

jimallman:
I've added to the wiki page, with details on [how we currently manage expired information](https://github.com/OpenTreeOfLife/opentree/wiki/Caching-results-from-common-API-calls#clearing-stale-cache-entries).

jar398:
Looks very thorough. Thanks

PR #591, opened 2015-02-07
Modify webapp config to use cached tree-view JSON

jimallman:
This assumes the corresponding support in phylesystem-api.
Addresses #494.


jimallman:
This is working now on **devtree**. The speedup is real, if not as dramatic as hoped.

jimallman:
Note that this depends on the deployment of [caching support](https://github.com/OpenTreeOfLife/phylesystem-api/commit/64242a5d26cfc64d26a85e0407c4c6291f8f2d9a) in **devapi** (or **api** on production).

PR #593, opened 2015-02-09
Simple caching (and clearing) of treeview comments.

jimallman:
This caches local comments indefinitely, but with explicit clearing
triggered by a webhook in the 'feedback' repo, defined here:
  https://github.com/OpenTreeOfLife/feedback/settings/hooks

This webhook is triggered by the 'Issues' and 'Issue comment' events; as
a result, adding/removing/updating/replying to an issue will clear all
cached comments. Once I have a better grip on the payload from GitHub
API, we should be able to selectively clear only the comments for the
affected location(s).

### DIFF
--- a/webapp/controllers/plugin_localcomments.py
+++ b/webapp/controllers/plugin_localcomments.py
@@ -664,7 +664,7 @@ def clear_matching_cache_keys(key_pattern):
     pprint("  %d items removed" % (item_count_after - item_count_before,))
 
 @cache(key=build_localcomments_key(request), 
-       time_expire=None, 
+       time_expire=60*5, 
        cache_model=cache.ram)
 def get_local_comments(location={}):
     # Use the Search API to get all comments for this location. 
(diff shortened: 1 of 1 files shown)


jimallman:
It seems we get expected notifications from the webhook when a new issue (or issue comment) is created, but not when one is modified or deleted. I can explicitly clear cache after deletion from the curation UI, but I'll need to tinker some more...

jimallman:
More work needed here:
- add explicit cache-clearing when deleting a comment from the curation UI
- troubleshoot faulty cache-clearing with root-relative URL (each attempt seems to clear several unrelated cache entries, but miss the one we want)

jimallman:
OK, this seems to be working nicely now on **devtree**, albeit without notification if someone on GitHub changes or deletes an issue or comment. To avoid confusion (but still provide some caching benefit under load), comments are now only cached for 5 minutes.

PR #597, opened 2015-02-13
Explicit routing for /cached/* URLs

jimallman:
This will fix the routing of clean & simple URLs for cached API calls,
even when phylesystem-api is not the default app in web2py. This fixes
a bug that was revealed by the different configuration of the [devapi](https://github.com/OpenTreeOfLife/deployed-systems/blob/5a4aa88944b88a92ebb15bd8e27f5cda406c498e/development/devapi.config#L41) vs.
[api (production)](https://github.com/OpenTreeOfLife/deployed-systems/blob/master/production/api.config#L45) system. Addresses #591.

### DIFF
--- a/deploy/setup/apache-config-shared
+++ b/deploy/setup/apache-config-shared
@@ -88,6 +88,12 @@
     RewriteRule ^/v2/study/(.*) /phylesystem/v1/study/$1 [PT]
 
     # ------------------------------------------------------------
+    # Explicit routing for /cached/* URLs (in phylesystem-api)
+
+    RewriteEngine on
+    RewriteRule ^/cached/(.*) /phylesystem/default/cached/$1 [PT]
+
+    # ------------------------------------------------------------
     # v1 API retained for transition period (ending Nov 2014 ?)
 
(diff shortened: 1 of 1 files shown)


jimallman:
This is tested and working now on **devtree**, including recent modifications to the devapi configuration to use [the welcome webapp](http://devapi.opentreeoflife.org/) as its default. Which is a little weird in my opinion, as it shows a "hello world" app at the root of our API server. The previous configuration showed a JSON entry point for exploring the `phylesystem-api` (see attached).

screen shot 2015-02-12 at 10 06 46 pm

jimallman:
FYI - I've also [added HTTP verbs to the cache keys in phylesystem-api](https://github.com/OpenTreeOfLife/phylesystem-api/commit/2a05b8333ad022189f713d9532ac15c3401b2c50) to distinguish responses to OPTIONS vs. POST.

jar398:
There should be no expectations about what the default webapp is... consider e.g. the asterales system where there is only one server, and the browser/curator runs on same machine as the api

Issue #555, opened 2015-01-20
Add tree metadata fields to curation UI (Tree > Properties popup)


jimallman:
As we're gathering per-tree metadata, we'll want to add (or choose) properties in Nexson to hold them. These will need to be included in the curation UI, possibly as controlled vocabularies.

jimallman:
Related to #552.

jar398:
Discussed on Feb 20th software call. Decision was to have field of checkboxes for inference methods and tree types ('check all that apply') together with (not as alternative to) a free form text input field for detailed and/or additional description.

jimallman:
All this implies a few changes to Nexson structure:
- remove `ot:curatedType` field probably (too vague)
- add `ot:inferenceMethods` (allows multiple values)
- add `ot:treeType` (just one value? or multiple?)
- add `ot:treeDescription` (or maybe just `description`?)

**How should we migrate existing studies in phylesystem?** We might sweep all the studies in phylesystem and...
 - create any missing fields
 - move existing `ot:curatedType` value based on value
    - if it matches one of our preferred inference-method options, move it to `ot:inferenceMethods`
    - if it matches one of our preferred tree-type options, move it to `ot:treeType`
    - any other text moves to `ot:treeDescription`
    - watch for near misses (e.g. "genes tree") and alert when they're found, for a closer look?

Of course, we'll also need to sweep the codebase for any references to `ot:curatedType` and change it to handle the new schema. Thoughts?

jimallman:
See related discussion in #552.

Issue #552, opened 2015-01-15
Nexson specification of tree inference method


josephwb:
The curator calls the tree inference method ```"Tree type"```. [Nexsons](https://github.com/OpenTreeOfLife/phylesystem-api/wiki/NexSON#table-i-predicate-vocabulary) call this ```"ot:curatedType"```. 

The nexson property name should be changed to be more informative.

Posting this here since it involves multiple tools: the curator, UI, and validation.

jar398:
Since it conveys the inference method, maybe the property should be called
ot:inferenceMethod, and the curator app should present it as "Inference
method". Just sayin'.

mtholder:
Just chasing down the history: the earliest reference I found to this was in this tread: https://github.com/OpenTreeOfLife/phylografter/issues/86
the notes ( https://docs.google.com/document/d/1tVXWaq3kqctlCh3qonkXxCphLeEJi0_k-3kSx949sQE/edit#heading=h.wk263bxa5r6t ) from the Oct 9, 2013 which @pmidford  alludes to don't specifically mention that tag, but do refer to a free-form text field describing the tree type.

jimallman:
> Since it conveys the inference method, maybe the property should be called ot:inferenceMethod, and the curator app should present it as "Inference method"

I think the problem is that it _may_ contain other information, at least in the studies from phylografter. In OpenTreeOfLife/phylografter#86, @pmidford says it has historically held other information:

> From Oct 9 call
> ot:curatedType (tree) - phylografter's type column - frequently contains inference method, but also other free text.

Maybe we should check for other values in the current phylesystem studies.

Hm, looks like the curation UI is actually inconsistent in its labeling. Here are a tree's properties when we're editing:

![screen shot 2015-01-19 at 9 34 46 am](https://cloud.githubusercontent.com/assets/446375/5802002/c01339c2-9fbe-11e4-9408-69abb2975143.png)

...and here's the same tree when we're just viewing the study (note the label for this field!):

![screen shot 2015-01-19 at 9 35 04 am](https://cloud.githubusercontent.com/assets/446375/5802004/c8a2eb3c-9fbe-11e4-88be-4875d647a35a.png)


I'll make a separate issue for this...

pmidford:
Perhaps we should add ot:inferenceMethod, the curator UI seems pretty 
clear about what should be there.  It would be nice to enforce a 
controlled vocabulary, but that is probably premature. Leave ot:type 
[...]

PR #598, opened 2015-02-13
Improved tree properties

jimallman:
These changes reflect our latest thinking in tree properties, as discussed in #552 and #555. Available now for review on **devtree**.

This also does a one-time capture of the deprecated `ot:curatedType` field before discarding it, as described in https://github.com/OpenTreeOfLife/opentree/issues/555#issuecomment-74102857. (This will have no effect if this migration has already been done on the server, which is probably the wiser path.)

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -59,6 +59,20 @@ var API_remove_file_DELETE_url;
 // working space for parsed JSON objects (incl. sub-objects)
 var viewModel;
 
+// Enumerate the expected values for tree inference methods. 
+// N.B. Multiple values are allowed!
+var inferenceMethods = [
+    // add a 'label' attribute if not the same as 'value'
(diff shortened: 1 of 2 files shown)


jar398:
Awaiting review from KC and the MIAPA list

jimallman:
@kcranston: The changes here to Nexson tree properties (incl. multiple **inference method** values, and a separate **tree type**) were the result of a lot of conversation, so I want to confirm that we're **not** planning to merge these changes. (As [I noted elsewhere](https://github.com/OpenTreeOfLife/opentree/pull/626#issuecomment-96894479), these commits also crept into the 'support-tree-collections' branch, so I'll need to reverse them in that PR.)

As a reminder, here's the current UI on production (before these changes):
screen shot 2015-09-19 at 4 16 26 am

And here's the UI as modified by this PR:
screen shot 2015-09-19 at 4 17 16 am

PR #626, opened 2015-04-28
Support tree collections

jimallman:
~~_**NOTE:** This branch is a work in progress and **should not be merged** in its current state._~~

This is a companion PR for the same-named branches of peyotl and phylesystem-api. It has basic support for tree collections (and some preparation for "favorites"), plus a new profile page for each curator where these will be shown.

This repo also contains our deployment tools and configuration templates. These reflect the addition of new document types and apache routing for them (which folllows our 'v2' conventions).


jimallman:
It looks like there are some unrelated commits in here (involving inference UI, and possibly stats?). Perhaps this should be rebased on another branch... I'll try to sort this out before we've finalized the new UI in this branch.

Issue #646, opened 2015-05-29
references page says it shows rank order, but doesn't


jar398:
https://tree.opentreeoflife.org/about/references says that the order of the studies is synthesis rank order, but the call that it uses to get the order (getSynthesisSourceList) is not documented to return in rank order, and as far as I can tell there's no reason it should.

jimallman:
The order shown here actually reflects the `findAllStudies` method, which I [mistakenly believed](https://github.com/OpenTreeOfLife/opentree/blob/6284a9ab4eb4c702871c2620c9d8fc65525d5dbd/webapp/controllers/about.py#L129) orders its results by priority. A [closer look at the code](https://github.com/OpenTreeOfLife/oti/blob/1e4538ba189ffe828e3f7733df3284937e364bb9/src/main/java/org/opentree/oti/plugins/QueryServices.java#L22-L40) does not support this.

The `/v2/tree_of_life/about` API method is simpler and better organized, and would be a good place to impose this order.

jimallman:
It seems we have no programmatic source for ordering trees by rank. And the References page is study-oriented, not tree-oriented, so it's not well suited for showing tree rank in its current form. For now, I'll remove the sort options from the References page so we aren't making promises we can't keep.

kcranston:
Agreed. Just provide an alphabetical-by-first-author list.

On Fri, May 29, 2015 at 1:10 PM, Jim Allman <notifications@github.com>
wrote:

> It seems we have no programmatic source for ordering trees by rank. And
> the References page is study-oriented, not tree-oriented, so it's not well
> suited for showing tree rank in its current form. For now, I'll remove the
> sort options from the References page so we aren't making promises we can't
> keep.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/OpenTreeOfLife/opentree/issues/646#issuecomment-106873707>
> .
>



-- 
~~~~~~~~~~~~~~~~~~~~~~~
karen.cranston@gmail.com
@kcranstn
~~~~~~~~~~~~~~~~~~~~~~~

jimallman:
I'm leaving this open, since it sounds like we plan on providing tree-rank data. Once this is available, I'll revisit the changes above and restore the sorting UI.

kcranston:
I think we can close this issue. The page no longer mentions rank ordering, so the issue is fixed. The question of how to best provide the ranked list of inputs for each version of the synthetic tree should be a separate issue.

PR #647, opened 2015-05-29
Remove faulty priority sorting in References page.

jimallman:
The server will now provide the default and only sort, alphabetically by
study's primary author. Addresses #646.

This is [working now on **devtree**](https://devtree.opentreeoflife.org/about/references).


kcranston:
Thanks! Merging.

PR #648, opened 2015-05-30
Sort studies by author (fwd or reverse)

### LINKED ISSUE #625: Sorting by author in curator
I have been rooting out some duplicate studies, and would find it helpful to be able to sort studies in the curator app by author name. I think would be of general use as well.

jimallman:
Note that this will try to use str.localeCompare() in newer browsers, so sorting of names with diacritical marks, etc. should conform to the user's expectations. Fixes #625.


kcranston:
looks fine to me. Merging.

Issue #670, opened 2015-07-09
Allow curator to answer licensing questions when importing from TreeBASE


kcranston:
At the moment, when you import a study from TreeBASE, all of the questions about license are greyed out and we automatically apply a CC0 waiver. We don't have the authority to do this and need to let the curator / author answer the questions about whether they can apply CC0.

jimallman:
According to [this Javascript comment](https://github.com/OpenTreeOfLife/opentree/blob/2b4276cc1eaf97bd556216e9cc4a16238aec374d/curator/static/js/study-creation.js#L130), I mistakenly thought the CC0 waiver was already applied to all TreeBASE data. I'll change this to prompt for a decision.

jimallman:
Just for background, see these related discussions (regarding implied waivers for TreeBASE data):
- https://groups.google.com/forum/#!topic/phylorf/1ewWPODLjGY
- http://sourceforge.net/p/treebase/mailman/message/29184870/ 

The current study-creation UI makes license/waiver options subordinate to the choice of import method (import from TreeBASE, vs. upload data), which was a deliberate change as we simplified this page. I'll need to reverse this, so that license/waiver options are always active regardless of the source/method chosen.

jimallman:
It seems the phylesystem API also assumes that studies from TreeBASE should always have CC0 waiver applied. This submission, for instance, specified the CC-BY-3.0 license, but the result was a new study with CC0 waiver applied:
```
curl 'https://devapi.opentreeoflife.org/phylesystem/v1/study' \
  -H 'Content-Type: application/x-www-form-urlencoded; charset=UTF-8' \
  -H 'Accept: application/json, text/javascript, */*; q=0.01
  --data 'import_method=import-method-TREEBASE_ID&import_from_location=IMPORT_FROM_TREEBASE&treebase_id=123&publication_DOI=&chosen_license=study-data-has-existing-license&cc0_agreement=false&alternate_license=CC-BY-3.0&alt_license_name=&alt_license_URL=&author_name=Jim+Allman&author_email=jim%40ibang.com&auth_token=de4916f3b38a737c2e90310db43df17760cb252b' --compressed
```
So I'm updating the creation UI, but the API will also need revision...

jimallman:
These fixes are live and available for review on **devtree**:  https://devtree.opentreeoflife.org/curator/study/create

[...]

PR #672, opened 2015-07-10
Always prompt for license/waiver for new studies

jimallman:
This addresses #670. We'd been assuming that studies imported from TreeBASE would always
have the CC0 waiver applied, but this is not the case.

NOTE that this changes the creation UI, but the underlying phylesystem API also needed a change! Look for a matching PR from branch `always-prompt-for-new-study-waiver` in **phylesystem-api**.


kcranston:
We decided to modify the text of the licensing options when data is coming from TreeBASE:

* option #2: “These data already have an applied waiver or license (highly unusual for TreeBASE data)”
* option #3: “These data have no applied waiver or license, and I have no authority to apply a waiver (usually true for TreeBASE data)"

Issue #551, opened 2015-01-15
Implement watch lists


jar398:
That is, a curator can elect to 'watch' a particular study (or collection...) which means they get email every time the study is changed (i.e. new commit).

Many implementation details to be worked out obviously.

kcranston:
Is this different from the favourites functionality mentioned in issue #675 ?

jar398:
If something is on a watch list, you get a notification when it changes.

Suppose you add a study, and then three months later someone makes an edit.
Just as with wikipedia, if you care about your curation effort, you care
when someone starts to muck with it, and checking it periodically is not a
workable approach.

This would be one line of defense against vandalism, but its interest goes
beyond that.

kcranston:
I understand the functionality that you are describing. Just wondering if we are talking about the same feature with favourite / watch / star (i.e. we have an issue that requests `favourite` and one requesting `watch` and it seems unnecessary to have two different buttons for the same or very similar functionality).

jimallman:
Yes, my mockups to date assume a single favorites feature that would handle both "bookmarking" collections (for easy checking later) and notification if desired.

josephwb:
+1 In addition to notification, it would be nice to see a summary of what has changed. At the moment the "History" tab in the curator is not all that useful (especially if a poor commit message was used, or none at all). Maybe, in addition to including such information in a notification email, the "History" tab could link to the git diff itself. For example, @snacktavish sent me [this](https://github.com/OpenTreeOfLife/phylesystem-1/commit/5817ed1ecc98b762652fcd68fbed9eceb0773ba3) so that I could see what she did.

jimallman:
> Maybe, in addition to including such information in a notification email, the "History" tab could link to the git diff itself. For example, @snacktavish sent me this so that I could see what she did.

Ah, great minds think alike. We can certainly build these diff links and add them to the History tab.

PR #692, opened 2015-08-13
Link to study diffs

jimallman:
This provides a detailed view (via GitHub's "diff" URL) for each entry in a study's History tab. This is a partial solution pending the development of a friendlier diff view, as discussed in #551.

Note that this depends on API support in the same-named branch of phylesystem-api, but it will fail gracefully if this is not deployed.


jimallman:
This is available for review now on **devtree**, in the History tab of the study curation tool.

PR #697, opened 2015-08-13
Fix for broken synth-tree login

jimallman:
We call the existing (JS) function [**fixLoginLinks**](https://github.com/OpenTreeOfLife/opentree/blob/96d25f2d9e87bd0d71b05d04589a5bbedd4d11a8/webapp/static/js/treeview.js#L596-L614) whenever the browser history changes. In the synth-tree viewer, we were calling this prematurely (before the Login link was created); this needs to be called again whenever a new Login link is created along with our comment form.

### DIFF
--- a/webapp/controllers/plugin_localcomments.py
+++ b/webapp/controllers/plugin_localcomments.py
@@ -38,6 +38,7 @@ function delete_all_forms() {
     });
 }
 function capture_form() {
+    // bind and modify widgets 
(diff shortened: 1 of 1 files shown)


jimallman:
This fix is working properly on **devtree**. Due to the severity of the bug, and since our normal PR reviewers are not available on IRC, I'm breaking protocol by merging + deploying this now.

PR #709, opened 2015-08-20
Add simple study validation on client side.

jimallman:
This will block attempts to save a study that won't pass the server-side validator. 

screen shot 2015-08-19 at 11 00 01 pm

Currently this is just a check for a non-zero integer in Metadata > **Study Year**. If we add more checks to the server-side validator, we should anticipate them here for friendlier error message.


jimallman:

PR #727, opened 2015-08-29
Use proper leaf count in synth-tree viewer

jimallman:
Now that [arguson provides a proper descendant-tip count](https://github.com/OpenTreeOfLife/treemachine/pull/191) (thanks @josephwb!), use it to improve node display and offer sensible Newick-download options. Some highlights:

- actual leaf nodes are now quite small and dark in color
- "visual" leaves (which have unseen descendants) remain hollow
- node size now reflects the _actual_ (vs _visible_) number of descendants, using a logarithmic scale

For the Newick download feature, I've set a threshold value (see `maxTipsForNewickSubtree` in `treeview.js`). Nodes with fewer descendant tips will offer to download the Newick subtree; other nodes will link to the full download page. Currently `maxTipsForNewickSubtree` is set to 1000 nodes; thoughts on a sensible value?

This is working now on **devtree**, see for example this view which has nodes above and below the current 1000-tip threshold: 
https://devtree.opentreeoflife.org/opentree/argus/opentree3.0@2713014


jimallman:
I also noticed that we were still offering EOL links to some unnamed nodes, due to multiple naming conventions used for internal nodes. We're now more careful to suppress this feature for '(untitled node)' and the like.

kcranston:
Now that we can get ntips, we want to provide a link to download the newick when ntips < threshold and a link to the full download on files when ntips > threshold. We also want to remove the depth limit (which was a hack for when we didn't have ntips). Sending this one back for that change.

jimallman:
@kcranston, the commits above (visible on **devtree**) should address all the issues in today's PR review, except for choosing a more informed value for the `maxTipsForNewickSubtree` (currently set to 10,000, see below). 

I've tested several requests to the low-level treemachine API on production, at a quiet time in the evening. Here are some results.

I started with [Aves](https://tree.opentreeoflife.org/opentree/opentree3.0@788774/Aves) (OTT id: 81461, node id: 788774), which has ~14,500 descendant tips. For a reality check on dev vs. production speeds, here's a test on dev (using its node id on dev):
```bash
$ time curl -v http://devapi.opentreeoflife.org/treemachine/v1/getDraftTreeForNodeID \
   -H 'Content-Type: application/json' -d '{"nodeID":696701}'
...
real	0m4.310s
```
... and the same test on production (much faster!):
```bash
$ time curl -v http://api.opentreeoflife.org/treemachine/v1/getDraftTreeForNodeID \
   -H 'Content-Type: application/json' -d '{"nodeID":788774}'
...
real	0m1.668s
```
Is there any difference in fetching a subtree based on OTT id versus node id? Let's find out:
```bash
$ time curl -v http://api.opentreeoflife.org/treemachine/v1/getDraftTreeForottId \
   -H 'Content-Type: application/json' -d '{"ottId":"81461"}'
...
real	0m1.486s
```
Not surprisingly, these take about the same length of time. Does this seem reasonable, given that we should expect more delays with multiple simultaneous users? 

Please give this a spin on devtree. For now, I've setting the threshold value to 10,000 descendant tips. Just keep in mind that devapi is probably about 3x slower than production. [Here's a node](https://devtree.opentreeoflife.org/opentree/argus/opentree3.0@2600268/Passeriformes--Sagittariidae--) that's just above this threshold, so you can experiment with its children.

jimallman:
I'll test again with the [newer v2 API methods](https://github.com/OpenTreeOfLife/opentree/wiki/Open-Tree-of-Life-APIs#subtree-tree-of-life), in case they're faster.

jimallman:
OK, we're now using the v2 APIs on **devtree**. Performance and output look good to me.

webapp/static/js/treeview.js
```diff
@@ -1113,48 +1117,38 @@ function showObjectProperties( objInfo, options ) {
     var subtreeDepthLimit = 4;
```
kcranston: this is no longer necessary, right?

Issue #750, opened 2015-09-15
file download link simply re-loads page


kcranston:
On the Files tab, the filename is a link that we claim you can (click to download). However, clicking on these links simply re-loads the Files tab. Example here:

https://tree.opentreeoflife.org/curator/study/view/ot_99/?tab=files

kcranston:
Also (and this should probably be a separate issue, but could probably be fixed together) the size should never be 0, right? It is in this example.

jimallman:
Yes, these bugs are close together (let's keep just one issue). The failures above are due to missing data in older supporting files, but in these cases I'll show something more sensible:
 - remove the hyperlink from a file with no `@url`
 - show an em-dash if there's no `@size` found

jimallman:
It looks like there's another problem in some studies, including your example above, where the supporting file is listed twice in the Nexson: once [for the study as a whole](https://github.com/OpenTreeOfLife/phylesystem-1/blob/master/study/ot_99/ot_99/ot_99.json#L106-L113), and again [under the tree for which it's a source file](https://github.com/OpenTreeOfLife/phylesystem-1/blob/master/study/ot_99/ot_99/ot_99.json#L951-L959).

In these cases, it seems only the second annotation is getting proper values for `@url` and `@size`, and the curation UI is checking the "wrong" annotation and finding neither. I'll chase this down...

jar398:
(While we're on the topic, remember I'm sitting on a bunch of supporting
files that had been deleted by accident... they need to be put back in
place)

jimallman:
Regarding duplicate annotations for these files, I've created a separate issue #752 to track this (larger) problem.

jimallman:
The original reported symptom is now resolved by PR #753 (current working on **devtree**). OK to close this? Or would you like to keep it open as a reminder of [@jar398's pending file restoration](https://github.com/OpenTreeOfLife/opentree/issues/750#issuecomment-140524570)?

jar398:
No need from my end to keep it open as a reminder about those files - I
worry about them quite frequently

jimallman:
> I worry about them quite frequently

OK, let me know if you'd like some help replacing them.

Issue #693, opened 2015-08-13
Comments in curator not recognized as edits


josephwb:
Comments of the front page of a study in the curator are not registered as an edit, so can't save when it is the only change to a study.

jimallman:
@josephwb: I gather this refers to the "Curator notes" in the Metadata tab? Will try to reproduce...

jimallman:
This is fixed now (working on **devtree**) and will be deployed to production as part of pull request #753. Please feel free to re-open this if not satisfied.

PR #753, opened 2015-09-16
Fix supporting-files display (and move old/local annotation messages)

jimallman:
This started as a set of cosmetic changes intended to improve the supporting-files UI (Files tab in curation app) when we have incomplete file information (see #750).

I've added more code to migrate (move or merge) any old "local" annotation messages to their proper homes inside annotation events of the Nexson `nexml` element. This code works very carefully, only moving those messages it can understand; if any migrations fail, the old message collections are preserved so we can try again later.

I've also added triggers to the file-description **and** curator-notes fields (see #693), so that edits to either will enable the Save button in the curation app.

This is live on **devtree** and ready to merge IMO.

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -850,10 +850,6 @@ function loadSelectedStudy() {
             if (['^ot:messages'] in data.nexml) {
                 data.nexml['^ot:messages'].message = 
                     makeArray(data.nexml['^ot:messages'].message);
-            } else {
-                data.nexml['^ot:messages'] = {
-                    'message': []
-                }
             }
 
             // move any old-style messages to new location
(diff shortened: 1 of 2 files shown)


jimallman:
Regarding the migration of old ("local") messages, here are the actual transformations when saving two files with duplicate supporting-file messages. The [first study had one pair of duplicate messages ](https://github.com/OpenTreeOfLife/phylesystem-0/commit/3343335f7aba708f7942b6565b7418112f7229fa) (local and centralized) and the second study had [nine sets of duplicates, one for each tree's source file](https://github.com/OpenTreeOfLife/phylesystem-0/commit/f370e5ac9bd3d7f2fa6b14316ce74d4537f47dd3
).

@mtholder, I believe the resulting Nexson matches our latest annotation standards. Please let me know if anything looks amiss!

You may also notice some other tweaks to the Nexson (adding expected fields to older studies); these are older code that's running against these (very) old Nexson documents.

curator/static/js/study-editor.js
```diff
@@ -850,10 +850,6 @@ function loadSelectedStudy() {
             if (['^ot:messages'] in data.nexml) {
                 data.nexml['^ot:messages'].message = 
                     makeArray(data.nexml['^ot:messages'].message);
-            } else {
```
jimallman: No sense adding this "local" message store now that we're back to keeping all messages inside their parent annotationEvents.

PR #754, opened 2015-09-18
Fix pip (requirements.txt) bugs

jimallman:
I've tested these changes on the **asterales** system (ot17), and they clear up a few deployment headaches:

- Failure of `git_refresh` now calls a dramatic halt to deployment and shows possible remedies.
- Unused `requirements.txt` files were removed or consolidated.
- Manipulations of `requirements.txt` files were removed. (Could we use multiple venvs instead?)
- Removed 'distribute' from webapp `requirements.txt`, after testing fresh deployments in **asterales** without it.

See commit messages below for more details.


jimallman:
These changes are live now on **devtree**. (I re-deployed all components there, bit I did not attempt to do drastic "fresh start" testing as with **asterales**.)

PR #755, opened 2015-09-19
Comment crash

mtholder:


mtholder:
deployed on dev-tree and seems to fix the issue.

PR #756, opened 2015-09-19
redirect top to a static page

mtholder:


jimallman:
@mtholder, I've updated the static page to reflect the hot-fixes done on production (plus the inline JS and CSS). Now that we've made other progress in site performance, we should evaluate whether we still need this.

kcranston:
Closing this because no longer needed (but keeping the branch for future reference)

PR #763, opened 2015-09-23
Run a background process to clear old sessions

jimallman:
Each time we (re)start web2py, we should also launch a background helper that periodically clobbers expired sessions, every 5 minutes by default.


jar398:
Probably better to do this as a cron job, rather than as an ad hoc background process

jimallman:
Here's [the source for `sessions2trash.py`](https://github.com/web2py/web2py/blob/master/scripts/sessions2trash.py) for reference.

kcranston:
Created issue for cron job: #845 so merging this.

PR #792, opened 2015-10-09
Change the tree-upload form validation.

### LINKED ISSUE #542: Curator UI: add new trees broken when schema is NEXUS
this is very minor:
from a trees tab (e.g. http://tree.opentreeoflife.org/curator/study/edit/ot_237/?tab=trees ), if you:

   1. select the file to upload, then
   2. specify the schema as NEXUS

then the "Add New Tree" button remains inactive. The user has to select the file again to make the button clickable.

jimallman:
This was using the value of the input[type=file] widget as part of form
validation; as it turns out, the upload widget clears this after each
file is added, so it's not a reliable test. Instead, we'll actually use
the filename preview, since this is maintained by the widget. (In each
test, we're looking for a non-empty field value *or* element text.)
Fixes #542.

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -4163,7 +4163,10 @@ function updateNewTreeUploadForm() {
         readyToSubmit = false;
     }
 
-    var chosenFile = $.trim( $('#treeupload').val() );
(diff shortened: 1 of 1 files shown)


jimallman:
This fix is available for review on **devtree**.

PR #810, opened 2015-11-23
Web2py on https only

### LINKED ISSUE #804: Web2py problem pickling a session (watch for this)
@josephwb experienced this when trying to add a study on https://tree.opentreeoflife.org. Here's the ultimate cause, from the traceback in web2py's `errors/`:
```
  File "/home/opentree/web2py/gluon/main.py", line 479, in wsgibase
    session._try_store_in_cookie_or_file(request, response)
  File "/home/opentree/web2py/gluon/globals.py", line 1089, in _try_store_in_cookie_or_file
    return self._try_store_in_file(request, response)
  File "/home/opentree/web2py/gluon/globals.py", line 1096, in _try_store_in_file
    or self._unchanged(response)):
  File "/home/opentree/web2py/gluon/globals.py", line 1041, in _unchanged
    session_pickled = cPickle.dumps(self)
  File "/usr/lib/python2.7/copy_reg.py", line 84, in _reduce_ex
    dict = getstate()
TypeError: \'NoneType\' object is not callable'
```

Normally, this means that a session has un-pickleable elements within it. But it seems to be a known issue that can result from using our combination of apache, mod_wsgi, and multiple "instances" of web2py -- in our case, on ports 80 (HTTP) and 443 (HTTPS). See discussion here:

- https://groups.google.com/forum/#!topic/web2py/CSuT_CknD20
- http://comments.gmane.org/gmane.comp.python.web2py/109600
- https://www.mail-archive.com/web2py@googlegroups.com/msg124892.html

If this problem persists, I gather there are ways of more explicitly "separating" the instances of `WSGIDaemonProcess` in apache config. We can try to glean those from the support threads above.

jimallman:
This is live on **devtree** and **devapi** and available for testing. So far, it's behaving as expected:

- all HTTP requests are redirected (with **301 - permanent redirect**) to HTTPS
- **wget** follows the redirect automatically
- **curl** apparently needs specific hint (`-L` or `--location` switch) to follow seamlessly
- python **requests** follows the redirect automatically

Based on anecdotal reports, this might fix #804 (failed logins due to broken session objects), since this appears to be caused by multiple web2py instances when using apache+wsgi.


jar398:
This is an incompatible change to the previously advertised http: URLs for the API and our users will have to be alerted. Not that that is a deal killer, but we need to take note of this.

Use of https: URLs is unaffected, and switching from http: to https: is the easiest migration path.
Use of http: URLs with GET (not POST) is unaffected (most tools will redirect automatically).
POST to http: URLs through the python 'requests' library is unaffected.
POST to http: URLs with 'curl' requires use of the -L --post301 curl flags.
POST to http: URLs through 'wget' only works with the very latest version of wget.

PR #819, opened 2015-12-14
Capture initial OTU mapping context in Nexson

jimallman:
This was updated in the UI, but not in the underlying Nexson model; as a
result, TNRS calls would use the wrong search context until/unless the
curator changed the context. Fixes https://github.com/OpenTreeOfLife/taxomachine/issues/120

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -7455,7 +7455,10 @@ function inferSearchContextFromAvailableOTUs() {
             ///console.log(">> inferredContext: "+ inferredContext);
             if (inferredContext) {
                 // update BOTH search-context drop-down menus to show this result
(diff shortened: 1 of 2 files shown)


jimallman:
_Copying this relevant comment from the taxomachine issue above:_

> It seems the initial mapping request is being submitted with the wrong context ("All life"), despite the UI clearly showing "Birds".

This has been happening during the first curation session for a study, before it's been saved in the curation app. When a study loads, we infer its search context by sampling taxa, but the updated UI (in this case, the selected "Birds") was not being captured in the Nexson. This is fixed now and working on **devtree**; see for example [study Foltz D., 2006](https://devtree.opentreeoflife.org/curator/study/edit/pg_1792/?tab=otu-mapping) (but please don't save this!).

Issue #300, opened 2014-06-09
Detect duplicate studies before import/creation, or in curation UI


jimallman:
This might be a belt-and-suspenders approach, with multiple checks:
 - curation UI checks when attempting to import/create a study (check TreeBASE id, DOI, reference string against oti)
 - curation UI checks when setting or changing DOI or ref-string of an existing study
      - warn only, in case it's an attempt to replace a prior entry that's hopelessly messed up?
      - or block this, so they have to delete the old/broken entry?
 - phylesystem-api (or peyotl validation?) should also check, in case someone is using the "raw" API instead of the curation UI

kcranston:
This came up again as we were thinking through the Dryad - OpenTree handshaking. If we can assume that OpenTree will never have more than one study with the same DOI, it makes the handshaking process much simpler. I wonder if we might revisit duplicates as an error rather than a warning?

jimallman:
>  I wonder if we might revisit duplicates as an error rather than a warning?

Or if the handshake is initiated from OpenTree, we can check for duplicates beforehand and block or warn the curator.

I thought we've heard some scenarios where the second copy of a study is preferred. I suppose blocking duplication entirely would force the submitter to review the existing study and either improve it or delete it. Any chance multiple curators would want to keep both around to debate their relative merits?

snacktavish:
I think no pros there. Eventually someone has to delete one of them, might as well just improve the existing copy from the get go, and block alternates.

jimallman:
> Eventually someone has to delete one of them, might as well just improve the existing copy from the get go, and block alternates.

Yes, this makes sense to me. I suppose if someone wants to import the entire study (maybe the original submitter didn't realize it was in TreeBASE), they can just clobber the original and start over.

jimallman:
This seems satisfied by PR #826. Regarding @kcranston's [comment about the Dryad handshake](https://github.com/OpenTreeOfLife/opentree/issues/300#issuecomment-92477090), this doesn't address the existing duplicate studies, nor will it block adding duplicates via our APIs.

PR #826, opened 2015-12-16
Disallow creation of duplicate studies

### LINKED ISSUE #808: Prevent duplicate check until some number of characters in DOI / URL box
When adding a new study, if you:
* select "I will upload tree data from my computer"
* click in the DOI / URL box
* navigate away from the page (for example, to look up a DOI)

You return to the page to find this:

screen shot 2015-11-19 at 2 45 25 pm

Should be waiting until there is some text there before checking duplicates.

jimallman:
This applies to both study creation and editing. Fixes #808.


jimallman:
This PR now fixes #300 (block adding a study with duplicate DOI).

PR #840, opened 2016-01-13
Curation app should use taxonomy API for MRCA test

### LINKED ISSUE #820: Use taxomachine 'lica' method instead of treemachine 'mrca' method
The treemachine mrca method should be reserved for finding mrcas in the synthetic tree. The functionality where it will optionally find mrcas in the taxonomy will go away when tm-lite is ready. So the curator app will need to be updated so that it uses the lica method from taxomachine to compute taxonomy mrcas.

Hmm, we may want to do something about excluding incertae-sedis-like taxa from the computation...

jimallman:
This uses [the `/taxonomy/lica` method from the v2 APIs](https://github.com/OpenTreeOfLife/opentree/wiki/Open-Tree-of-Life-APIs#lica).
Fixes #820?


jimallman:
This is working now on **devtree**, see for example https://devtree.opentreeoflife.org/curator/study/edit/pg_2816?tab=trees&tree=tree6556

Issue #737, opened 2015-09-02
Bulk selection of taxa on mapping page


josephwb:
For big trees there may be tips that do not currently have a satisfying mapping (but may, say, in the new taxonomy). When these are rejected, they return to the mapping queue when one selects "continue mapping". This is cumbersome when this list of taxa grows into the dozens or more, as either 1) they need to (manually) be rejected several times, or 2) they have to (manually) be unchecked each time a mapping procedure occurs.

What would be useful: some sort of bulk selection; say, click on one taxon, hold shift, and click on the last contiguous taxon that you wanted (un)checked. Currently the curator only allows (un)check all.

kcranston:
Related to #809

jimallman:
> What would be useful: some sort of bulk selection; say, click on one taxon, hold shift, and click on the last contiguous taxon that you wanted (un)checked.

At @jar398's prompting, I've looked at how Gmail handles this, and it's a nice simple approach: Keep track of the last item (de)selected in the current list. If the next item is (de)selected with the shift key held down, (de)select all items in this range to match. _N.B. that this means the *second* item is the one that dictates whether the items are selected or deselected._

Since checkboxes indicate a persistent selected state, this avoids any need for additional modifiers like `Command/Control` keys.

Any major change in list display (via pagination, filtering, etc) will clear the last-selected item. In this case, shift-clicking an item will toggle that item only, just as if the shift key was not pressed.

PR #855, opened 2016-01-21
Shift-select multiple OTUs in mapping tool

jimallman:
This fixes issue #737, following the simple interaction model used in Gmail (see [comments in that issue](https://github.com/OpenTreeOfLife/opentree/issues/737#issuecomment-173621771) for details). I've also added more visible highlighting of the selected OTUs, as shown here:

screen shot 2016-01-21 at 6 26 06 pm

(The slightly darker row *Capparis leprieurii* is the one under the mouse, which would be pale gray if not selected for mapping.)

This is working now on **devtree**. Comments welcome!

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -1242,6 +1242,9 @@ function loadSelectedStudy() {
                     }
                 });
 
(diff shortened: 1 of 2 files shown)


kcranston:
This is awesome! Merging.

PR #857, opened 2016-01-22
Trim whitespace from regex'd OTUs when mapping.

### LINKED ISSUE #815: Remove hanging whitespace when modifing OTU names
If a modified taxon name contains a hanging whitespace (for example, because of a regex modification), it will fail mapping. This is really hard to detect. Should strip these (or warn?).

jimallman:
Fixes #815.


jimallman:

Issue #833, opened 2015-12-28
Advance from draft3 to draft4


jar398:
Under 'latest synthesis release' we currently have https://tree.opentreeoflife.org/about/synthesis-release/v3  i.e. v3 when it should be v4

jimallman:
@jar398, I'm guessing this is due to a lack of statistics for this release. Will investigate further...

jimallman:
See [the manually-constructed file `synthesis.json`](https://github.com/OpenTreeOfLife/opentree/blob/master/webapp/static/statistics/synthesis.json), which will need an entry for v4.

jimallman:
@jar398, I've [added an entry for synth draft 4](https://github.com/OpenTreeOfLife/opentree/blob/988c71d05e6672120d84740c6e0c9a72c5beaead/webapp/static/statistics/synthesis.json#L26-L36), with sensible-looking information (sources noted in inline "comments"). 

Can you confirm this information and add sensible values for `tree_count`, `total_OTU_count`, and `tip_count`? Or remind me how I can determine them myself -- I haven't found any notes on how to do this with confidence.

jimallman:
From #864:
> Updated and added to development branch. Needs to be tested.

I've tested the links (on **devtree**) from the About menu, between synth-release versions, and between the taxonomy and synth-release pages. All are working as expected. Can we close this issue?

jar398:
Yes.

PR #864, opened 2016-01-23
Add initial stats for v4 synthetic tree. 

jimallman:
**N.B. these stats are not ready to merge to master, pending reliable counts.** I've deployed them as-is to **devtree**, mainly to show the effect on site navigation and working cross-links to OTT2.9, etc.

See inline comments for sources used. NOTE that this does not yet include proper tallies, which will need to be obtained from the data download (@jar398 ?). Addresses #833.

### DIFF
--- a/webapp/static/statistics/synthesis.json
+++ b/webapp/static/statistics/synthesis.json
@@ -24,14 +24,12 @@
         "tip_count": 2339460
     },
     "2015-10-24T00Z": {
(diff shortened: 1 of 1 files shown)


jar398:
Updated and added to development branch.  Needs to be tested.

It was surprising to me that the number of OTUs decreased by about 1/3 going from v3 to v4, from 63,393 to 41,029. I wonder why that was. Perhaps they were counted in different ways.

kcranston:
Merging, but also going to investigate drop in OTUs (separate issue). Will keep branch for now.

PR #866, opened 2016-01-29
fix v2 and v3 numbers

josephwb:
I don't know where those v2 numbers came from, but they are wrong. I have checked the DB and the values are now correct. Unfortunately, no value for `total_OTU_count` was ever computed (again, no idea where the previous value of 60933 came from).


jar398:
I've fixed the OTU count for v3 as well, and will change the title of the PR.

kcranston:
Merging, but leaving the issue open

Issue #769, opened 2015-09-28
bad OTU mappings - how to help prevent?


rgazis:
https://tree.opentreeoflife.org/curator/study/edit/pg_529/?tab=trees

All taxa in the tree - I believe- belong to Hypocreales but the MRCA test (both taxonomy and synthesis) give as a result "Leotiomyceta". There are some taxa outside the Hypocreales in the study BUT not in the tree. Does the MRCA test is for the taxa in the tree or takes into account all taxa included in the study (and present in other trees)?

![screen shot 2015-09-28 at 5 45 49 pm](https://cloud.githubusercontent.com/assets/6391539/10149725/547f90d6-6609-11e5-803e-96e2b40a6630.png)

Thanks,

Romina

rgazis:
Same thing in this study but at different level

https://tree.opentreeoflife.org/curator/study/edit/pg_1669/?tab=trees
![screen shot 2015-09-28 at 6 11 57 pm](https://cloud.githubusercontent.com/assets/6391539/10150284/a79ffdac-660c-11e5-8416-0be8eebacb84.png)

And the MRCA from the taxonomy test differs in one hierarchical level than the MRCA test results from the synthetic tree

![screen shot 2015-09-28 at 6 10 17 pm](https://cloud.githubusercontent.com/assets/6391539/10150262/9381ace4-660c-11e5-9ea7-2be0eebf0c37.png)

kcranston:
The tests should only be based on the taxa in the ingroup of this tree. I agree that it is difficult to diagnose where the problem might be, though. Wonder how we might make this easier?

rgazis:
That would make sense. But, then something is not working, all the taxa in this tree 

https://tree.opentreeoflife.org/curator/study/edit/pg_1669/?tab=trees&tree=tree5902
belong to Epichloe

jar398:
We should check to make sure that the taxonomy is right before spending a
lot of time staring at the treemachine method. There is some trickiness
trying to make Hypocrea and Trichoderma be the same genus (as requested),
and perhaps I got it wrong.
Will think about how to best do this...

jar398:
I don't think it's the taxonomy.  Here's what I get when I look up the
lineages of all the OTUs in that study:

Epichloe baconii < Epichloe < Clavicipitaceae < Hypocreales
Epichloe glyceriae < Epichloe < Clavicipitaceae < Hypocreales
[...]

Issue #719, opened 2015-08-26
Option to "Approve all perfect matches" only


josephwb:
The "Approve all" button on the mapping page apparently accepts all unique mappings, whether they are perfect or not (e.g. synonyms). I would like an option to accept all perfect matches only, as it is only the remainder that require scrutiny. :rocket:

This replaces the unclear issue https://github.com/OpenTreeOfLife/opentree/issues/718

josephwb:
If this is not going to happen, having the ability to sort by map type (perfect, synonym, etc.) would be helpful.

josephwb:
Another idea: when clicking on "Approve all", a window pops up asking what should be approved:
1) All mapped
2) Only perfectly mapped

This seems related to https://github.com/OpenTreeOfLife/opentree/issues/730.

rgazis:
Having the options would be a good idea. "All mapped" would accept all the labels colored in green, right? because in some cases I noticed that even the labels colored in blue (synonyms) are also automatically accepted.

josephwb:
No, my idea was that:

2) Only perfectly mapped

would be for the green matches, and 

1) All mapped

would be any non-ambiguous mapping (i.e. may map to a single (blue) synonym). I believe that this is the way the curator currently works (and what I would like changed).

jimallman:
For now, I'm modifying this behavior to be more restrictive, as requested in #730. (All synonyms, homonyms, and less-than-perfect matches will require manual review.)

PR #872, opened 2016-02-02
More careful auto-approval of OTU mappings

### LINKED ISSUE #730: "approve all and continue" also approve non-exact matches
Study
https://tree.opentreeoflife.org/curator/study/edit/ot_365/?tab=otu-mapping

Normally only exact matches are accepted when clicking "approve all and continue" (green and only one option for mapping).

In this case, the mapping options with a blue legend and a different name are accepted when clicking "approve all and continue".

P.S. I will not save the mapping in this study, so somebody can check if there is a bug

R

jimallman:
We now skip auto-approval for OTUs with one suggestion, if
  - it's a synonym
  - it's a taxon-name homonym
  - its matching score is less than perfect

I've also tweaked the auto-approval button text so it doesn't
over-promise, and added CSS to prevent line-wrapping for mapping
suggestions (so approval buttons stay alongside the suggested name).
Fixes #730. Addresses #719 and #769.

### DIFF
--- a/curator/views/study/edit.html
+++ b/curator/views/study/edit.html
@@ -1252,7 +1252,8 @@ body {
                                     <i class="icon-remove"></i>
                                 </button>
                             </div>
-                            <a class="btn btn-mini"
+                            <span style="white-space: nowrap;">
+                                <a class="btn btn-mini"
                                                   style="padding: 0px 4px;"
                                                   data-bind="click: approveProposedOTULabel">
                                 <i class="icon-ok"></i></a>&nbsp;<a data-bind="text: proposedMapping(otu)[0].name,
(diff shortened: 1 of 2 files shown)


jimallman:

curator/views/study/edit.html
```diff
@@ -1409,11 +1411,11 @@ <h4 style="margin-top: 0;" data-bind="visible: !autoMappingInProgress()">Mapping
                 <div class="mapping-batch-operations" style="display: none;">
                     <button id="batch-approve" class="btn btn-small pull-left"
                             onclick="approveAllVisibleMappings(); return false;">
-                        <i class="icon-ok"></i> <strong>Approve all</strong> and continue
+                        <i class="icon-ok"></i> <strong>Auto-approve</strong> (continue)
```
kcranston: would be better if the button text was "Approve exact matches"

jimallman:
Revised text "approve" => "accept" is visible on **devtree**.

Issue #809, opened 2015-11-20
Sort 'failed mapping' below unmapped in OTU list


kcranston:
When mapping taxa, the default sort order is 'Unmapped OTUs First'. The UI does not distinguish between 'unmapped' and 'failed mapping'. For long OTU lists, that means you eventually get a long list of failed OTUs at the top and have to scroll down to see unmapped.

kcranston:
Another way to deal with this is a filter that shows 'unmapped only', 'mapped only', 'failed mapping'.

jimallman:
> The UI does not distinguish between 'unmapped' and 'failed mapping'.

This is the crux of the problem. During mapping, we want each OTU to "stay put" so that the user can respond to the mapping choices offered for each. Making the change described above would cause failed OTUs to immediately drop, probably out of sight entirely. 

Maybe that's good. It would fill the list with meaningful choices, and as these are resolved, the failed OTUs would reappear for modification and retries. I'm tempted to try this behavior and see how it feels. Thoughts?

jar398:
From a user's point of view, distinguishing failed from not-yet-mapped
would be useful. If not-yet-mapped, it could easily be productive to spend
time trying to map, but if marked failed, they will know they (or a
previous curator) has tried to map and failed, so probably not worthwhile
to attempt to map. So the information saves them time.

Probably too hard to fix at this point, so just sayin'. And I don't know
how we'd represent this distinction in the nexson.

jimallman:
This is working now on **devtree**. See commit message above for details on the behavior, esp. that it doesn't leave a persistent "failed" attribute in an OTU's Nexson.

jar398:
Would making the flag persistent be a big pain?

jimallman:
Probably not a big deal, no. I think we just need to agree on a name (`^ot:failedOTUMapping`, perhaps?) and make sure it's not scrubbed or altered by peyotl or its validators. And I'll need to tweak the client-side code in a few places to make sure it's set and cleared at appropriate times.

jar398:
hmm, probably best to wait for peyotl to implement it before doing any
further work on it.  Thanks​

kcranston:
I am going to consider this fixed with PR #875

PR #875, opened 2016-02-02
Force failed mappings below other unmapped OTUs.

jimallman:
This should make it less frustrating for curators, as described in #809.
N.B. that this effect will not persist across sessions; clearing
(resetting) these mapping results will restore the "failed" OTUs to
their position among the other unmapped taxa.


jimallman:

Issue #873, opened 2016-02-02
When adding trees to collection, provide list of treeIDs once user enters study ID


kcranston:
The help text in the tree ID box indicates that the ID should be a bare number:

<img width="457" alt="studyid_treeid" src="https://cloud.githubusercontent.com/assets/312034/12766822/977bfe96-c9d3-11e5-986e-cc910f9a49c1.png">

But it actually requires 'tree' at the start of the ID. For example, for this study, the tree ID is `tree6620` not simply `6620`.
https://devtree.opentreeoflife.org/curator/study/view/pg_2850/?tab=trees

kcranston:
Noting that not all tree ids start with 'tree' (or any other constant text).

jimallman:
> An even better solution than PR #880 would be to grab the tree IDs once the user inputs the study ID. Changed title to reflect, and therefore leaving open.

I had originally intended this option for a user who has the study and tree IDs in hand already. Otherwise, I expected them to navigate to the study+tree in question and add it from there (the "cherry-picking" scenario).

Your suggested change is a minor project, but certainly feasible. If we want to make it easier to find studies+trees from any page, both the study-ID and tree-ID inputs should show a list of matching values as the user types. We could match on IDs, author name, publication year, tree label, etc. Thoughts?

kcranston:
The studyID is fairly easy to find, so having the user copy it (mentally or physically) from another browser tab seems ok to me. The treeID is harder to find, the format is less consistent, and having to copy two pieces of information is just more complicated. The 'search on anything' sounds great, but I think that other projects are higher priority. Effort estimate on auto-completing treeID given studyID? If a new oti query would help, do make an issue in the [ottreeindex](https://github.com/OpenTreeOfLife/ottreeindex) repo.

jimallman:
This is working now as of https://github.com/OpenTreeOfLife/opentree/pull/981, and available for review on **devtree**. It uses the 'search on anything' behavior for studies, then a drop-down list of trees for the chosen study.

PR #880, opened 2016-02-04
add 'tree' prefix to id placeholder

jar398:
Not tested yet... don't see what can go wrong, but if someone insists I'll try it out before going to production

### DIFF
--- a/curator/views/layout.html
+++ b/curator/views/layout.html
@@ -850,7 +850,7 @@ from applications.opentree.modules.opentreewebapputil import get_user_login, get
                 <label>and tree ID</label>
                 <input type="text" name="tree-id"
                     class="input-small" style="margin-bottom: 2px;"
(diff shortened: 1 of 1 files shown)


kcranston:
Where is this being added in the UI?

kcranston:
Ah, in the Add new tree to collection. Is the prefix always 'tree'?

jar398:
It's only the placeholder, so has no effect on what the webapp does. But it can affect what the user does, by encouraging them to put the full id there, such as tree1234 or Tr1234, instead of just the number 1234, which would be wrong.  We could put any prefix we like there, but we ought to put some prefix to avoid misleading people into thinking that an integer could ever be an id. This situation occurs elsewhere in the webapp and there it properly has a prefix.

kcranston:
Addresses issue #873

Issue #744, opened 2015-09-08
Collect a URL when user submits classification/phylogeny advice


jar398:
If the user selects 'Feedback on tree of life' then some effort should be made to collect from the user a URL of an authoritative source (preferably a DOI). This is a requirement we have on taxonomy changes, and the feedback given here may bear on the taxonomy, if it bears on phylogeny it's just as important.

Something like 'When proposing a correction or improvement to the tree, please provide the DOI or URL of a published source or database that documents the desired topology or nomenclature.'

jar398:
Here is an example of what I'm talking about:
https://github.com/OpenTreeOfLife/feedback/issues/101
The submitter provided a free-text reference, but no URL.  We really really want a URL, ideally a DOI, in these cases.

jar398:
of course they didn't use the feedback form for this submission since the feature is down now, but you get the idea.

jar398:
I wonder if we could increase the priority of this one. I've received probably 20 or 30 URL-less taxonomy corrections, and more are coming. Here's just one: https://github.com/OpenTreeOfLife/feedback/issues/153

kcranston:
Good idea. I can probably tackle this tomorrow.

jimallman:
@kcranston, are you working on this one? If not, I think I can knock it out pretty quickly.

jimallman:
One more question: Should this URL be a required field?

jar398:
Good question.  When the feedback is scientific advice, then I think yes, a
URL should be required. But other kinds of feedback, such as "it would be
nice if the labels were in Comic Sans", don't need a URL.

If someone really wants to dig in their heels and provide their advice
without evidence, they can provide a nonsense URL like about:blank or
http://google.com/.

jimallman:
> other kinds of feedback, such as "it would be nice if the labels were in Comic Sans", don't need a URL.

Certainly. I thought we'd hide or show this field based on the feedback type selected. (Don't we do some of this already?)

jar398:
Hmm.  There are kinds of taxonomy advice that do not require URLs, e.g. duplicates. So now I'm not so sure about what I said about absolutely requiring a URL and nonsense URLs. E.g. https://github.com/OpenTreeOfLife/feedback/issues/189
(I wonder why so many feedback issues have no flags associated with them?...)

jimallman:
> I wonder why so many feedback issues have no flags associated with them?

[...]

PR #889, opened 2016-02-08
Add supporting URL to phylo-feedback

jimallman:
**STILL IN PROGRESS**. This addresses #744; see that issue's comments for more detail and ongoing discussion. The work so far is available for review on **devtree**:

  - Added a non-required 'supporting reference URL' field for phylogenetic feedback
  - Removed lots of diagnostic chatter in web2py logs
  - Finished smarter cache-clearing for local comments (incl. when adding a new issue)

### DIFF
--- a/webapp/controllers/plugin_localcomments.py
+++ b/webapp/controllers/plugin_localcomments.py
@@ -396,12 +396,14 @@ def index():
 
         try:   # TODO: if not comment.deleted:
             # N.B. some missing information (e.g. supporting URL) will appear here as a string like "None"
+            supporting_reference_url = metadata.get('Supporting reference', None)
+            has_supporting_reference_url = supporting_reference_url and (supporting_reference_url != u'None')
             markup = LI(
                     DIV(##T('posted by %(first_name)s %(last_name)s',comment.created_by),
                     # not sure why this doesn't work... db.auth record is not a mapping!?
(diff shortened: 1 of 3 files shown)


kcranston:
Ok, now that we can actually see this... ;)

I agree with @jar398 on issue #744 that we should remove the second pulldown entirely. The single pulldown would have the following categories:

1. General feedback
* Correction to relationships in the synthetic tree
* Suggest a phylogeny to incorporate
* Correction to names (taxonomy)
* Bug report (website behavior)
* New feature request

and only the 2nd, 3rd, and 4th require URLs (and the help text should say URL / DOI). If someone wants to leave feedback without a URL, they can always choose 'General feedback'.

jimallman:
OK, the latest code (incorporating feedback above) is live on **devtree**.

jimallman:
Once this is merged and approved, I believe it will be fairly easy to consolidate all the labels in the _feedback_ issue tracker. We should be able to search by (old) label, select all, assign the appropriate new label (if any), and clear the old. Then clean up the label list to include only the new ones.

kcranston:
On gitter, @jimallman and I decided to include a "Please fill out all fields" and grey out the Post button unless all fields complete (otherwise, you get a series of popup warnings that say "you need to fill out field X", then one for "...field Y", then "...field Z")

Issue #891, opened 2016-02-08
Revise latest OTU mapping UI to avoid confusion


jimallman:
Currently, a curator "accepts" a proposed mapping (ie, marks it as correct) by clicking a button with a checkmark icon, as shown here:

<img width="787" alt="screen shot 2016-02-08 at 4 03 45 pm" src="https://cloud.githubusercontent.com/assets/446375/12899284/8c7fbcee-ce7e-11e5-9724-e37d496a3b22.png">

The same style button is used for single and multiple suggestions. Early users report some confusion because these squarish buttons look like checkboxes, so it looks like all options are already marked as chosen. To avoid this, we might choose another icon, of possibly dim/hide the icon in the button's default state, showing it only on mouse-over. Here are some options. Reactions welcome!

<img width="789" alt="screen shot 2016-02-08 at 4 01 27 pm" src="https://cloud.githubusercontent.com/assets/446375/12899347/e125a6be-ce7e-11e5-9641-0c4214c0ea8d.png">
_A) Empty button with a ghosted check on mouse-over._

<img width="792" alt="screen shot 2016-02-08 at 4 01 11 pm" src="https://cloud.githubusercontent.com/assets/446375/12899349/e12e1ede-ce7e-11e5-93c6-f559242f9d35.png">
_B) Empty button with an opaque check on mouse-over._

<img width="798" alt="screen shot 2016-02-08 at 4 00 41 pm" src="https://cloud.githubusercontent.com/assets/446375/12899348/e12d134a-ce7e-11e5-998f-bba442765a83.png">
_C) Ghosted button with an opaque check on mouse-over._

__Edited to correct order of A,B,C options.__

snacktavish:
I like A! and I feel fine about B and I think C is still a bit confusing.

hdliv:
I like A too, B is fine..but C still has all those checks that I find confusing

jimallman:
thanks! seeing no other obvious icons to use instead, i'll go ahead with option A. it's easy enough to test.

jimallman:
These changes (option A above) are available for review on **devtree**.

PR #892, opened 2016-02-09
Revise OTU mapping UI ("checkbox" buttons)

jimallman:
Make the buttons distinct from declarative checkboxes, as suggested in
issue #891. Also add tooltips to accept and reject buttons, for single
and multiple options.


jimallman:

PR #898, opened 2016-02-14
Adapt to new v3 APIs (and ArguSON)

jimallman:
~~WORK IN PROGRESS.~~ This branch attempts to properly parse and render the new ArguSON format. Until the new treemachine (tm-lite) is available, this can be tested by forcing a load of the [static test file `test-arguson.json`](https://github.com/OpenTreeOfLife/opentree/blob/adapt-to-new-arguson/webapp/static/test-arguson.json) instead of a changing target node in the synthetic tree. See [the `config.example` file](https://github.com/OpenTreeOfLife/opentree/blob/adapt-to-new-arguson/webapp/private/config.example#L52-L53) for an example of how to do this.

N.B. There are still a few pending changes to the new ArguSON. In some cases like `tax_sources`, this logic will [correctly handle the current and proposed solutions](https://github.com/OpenTreeOfLife/opentree/blob/adapt-to-new-arguson/webapp/static/js/treeview.js#L675-L695). Others will require new tweaks.


jimallman:
**UPDATE:** This is now pretty much the feature branch for migrating to v3 of the `tree_of_life` APIs (a.k.a. `tree_of_life`).

jimallman:
See remaining release tasks in https://github.com/OpenTreeOfLife/treemachine/issues/207, and specific methods used in webapps in... hm, can't find this issue(?) with checkbox list...

jar398:
I want a round of review for new-arguson before finalizing this. I'm
preparing a spec, should be ready in a couple of days.

webapp/static/js/argus/drawtree.js
```diff
@@ -1861,7 +1870,7 @@ ArgusNode.prototype.isActualLeafNode = function() {
     return this.num_tips === 0;
 };
 ArgusNode.prototype.isVisibleLeafNode = function() {
-    return this.hasChildren === false || this.num_tips === 0;
+    return this.hasChildren === false || this.num_tips_in_view === 0;
```
jimallman: Minor correction to my commit message. This is used to detect **visible** tips, not "true" tips from the synthetic tree.

Issue #854, opened 2016-01-21
add OpenTree citation to "About" page


arlin:
I went to the main web site just now, looking for a citation to the OpenTree paper.  I couldn't find one.  There is a "Bibliographic references" item in the "About" pull-down menu, but this has citations for a huge list of source studies.  

I think a link to Hinchcliffe, et al 2015 belongs on the "About: Open Tree of Life" page at https://tree.opentreeoflife.org/about/open-tree-of-life.

PR #900, opened 2016-02-16
Added citation reference of OTOL to About OTOL page.

coatless:
Addresses #854


kcranston:
w00t! Thanks for this.

PR #901, opened 2016-02-17
Tree conflict viz

jimallman:
Since we're talking about rolling this out soon, it seems like time for a proper pull request. I believe we have a few usability issues to work out, from small to large:

- [x] The summary report lists could use some work (hide opaque node ids, make each listed name a link to the taxo-browser?)
- [x] Show some kind of spinner or "wait" screen while fetching+displaying conflict?
- [x] Indicate to the curator that conflict will **always** apply to the _last-saved_ version of the current input tree, and not to any recent tweaks that haven't been saved. (How hard would it be to change this, and submit the tree JSON for faster iteration?)
- [x] We currently launch (from the summary report) a conflict-view of the input tree in a separate browser window. While this is a "view" URL, it's an open invitation to start editing. Editing a single study in two browser windows is an invitation to disaster, since the curation app is essentially a single-page app. Changes saved from window A will overwrite those from window B, and vice versa. No bueno. Instead, we should consider opening these links in the _same_ window (switch to Trees tab, etc) by default.

_**UPDATE:** Some of the items above have quick fixes, but not proper solutions. I've moved those concerns to new issues. [@jimallman]_

And of course we have some [ideas about better conflict display](https://gitter.im/OpenTreeOfLife/public?at=56c4e06b611729797fd285ca)...


kcranston:
Is there a way to give a warning / error if someone opens a duration session if they already have an open session?

I don't like the same window idea - I've already found it useful to have the report and tree open side by side.

jimallman:
> Is there a way to give a warning / error if someone opens a duration session if they already have an open session?

No, there's currently no safeguard in place for this, other than our solution for two curators working independently: accept all saved changes, attempt to merge them automatically, and complain if there's a conflict. 

I'll ponder whether there's something we can do that's smarter than this, even just for a single curator. Perhaps we try to keep track of whether someone is editing a study, and allow only one "edit" page open at a time. Other windows would open as "view" only, and block (or warn strongly) if they click the Edit button. Alas, this will always be a guessing game based on session expiration, timeouts, etc.

jimallman:
For now, I'm planning to address the issues above with simple (simplistic?) fixes:
- a "spinner" while conflict info is being fetched for display (plus smart "dimming" of its UI)
- tweaks to the summary report list, as described above
- a prominent reminder about last-saved study being the basis for conflict analysis (perhaps using our "unsaved changes in Nexson" test to determine whether this should be shown)

PR #908, opened 2016-02-19
Tree conflict viz (ongoing improvements)

jimallman:
**WORK IN PROGRESS**. I've revived this feature branch for ongoing tweaks to conflict reporting and visualization. As a first step, I've added explicit control (from JS) for whether to re-use cached responses (for a given set of trees) from the conflict service [as discussed on Gitter](https://gitter.im/OpenTreeOfLife/public?at=56c63afd833e58f50fb0082a).

### DIFF
--- a/curator/private/config.example
+++ b/curator/private/config.example
@@ -62,7 +62,7 @@ findAllStudies_url = {oti_domain}/findAllStudies
 singlePropertySearchForStudies_url = {oti_domain}/singlePropertySearchForStudies
 singlePropertySearchForTrees_url = {oti_domain}/singlePropertySearchForTrees
 findAllTreeCollections_url = {opentree_api_domain}/../../v2/collections/find_collections
-treeConflictStatus_url = {conflict_api_domain}/conflict-status?tree1={TREE1_ID}&tree2={TREE2_ID}
+treeConflictStatus_url = {conflict_api_domain}/conflict-status?tree1={TREE1_ID}&tree2={TREE2_ID}&use_cache={USE_CACHE}
 
 # Open Tree API - RESTful URLs for managing studies in remote storage (note HTTP verbs for each)
 #
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -1787,7 +1787,7 @@ function displayConflictSummary(conflictInfo) {
     }
 }
 
-function fetchTreeConflictStatus(inputTreeID, referenceTreeID, callback) {
+function fetchTreeConflictStatus(inputTreeID, referenceTreeID, callback, useCachedResponse) {
     // Expects inputTreeID from the current study (concatenate these!)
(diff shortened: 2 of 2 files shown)


jimallman:
@jar398, @kcranston: As discussed, all taxonomically-mapped leaves of an input tree will show support in the conflict viz, including appropriate links to the taxo-browser (or synth-tree viewer, depending on our analysis). This is working now on **devtree**. I like it!

screen shot 2016-03-09 at 1 15 33 am

jar398:
This is great, and OK as is. But I have two nits:

1. why the 'aligned with' only appears after you've done a conflict
analysis; I don't see any reason not to provide the link before a conflict
check has been requested.

2. I think the popups should look the same for tips and non-tips in showing
the nexson node id (either in both cases or neither case). The taxon name
in the first line is redundant with the taxon name in the third line, so
this would not lose any information, and nexson node ids for tips would be
at least as useful as nexson node ids for internal nodes, in my opinion
(both are mainly of interest for people using the API or parsing nexsons).
​

jimallman:
> I don't see any reason not to provide the link before a conflict check has been requested.

I think the link might be useful, but the verbiage "Aligned with" seems odd without surrounding conflict information. Oh, and which of the two links would be the default: one to the OT taxonomy browser, or to the synth-tree viewer?

jar398:
Sure. Since we talk about "OTU mapping" then maybe the verbiage could be
"Mapped to". And the default link would have to be to the taxonomy browser
(since the node may not be present in the synthetic tree, and also that's
what you get during OTU mapping).

I have wanted this feature a number of times, by the way, so not just
trying to make work.

kcranston:
Thanks, @jimallman ! I agree with @jar398 about including node IDs, and with removing redundancy (name multiple times in popup). I think simply saying "Mapped to" instead of "Aligned with" would remove confusion.

jimallman:
> the node may not be present in the synthetic tree

I'm concerned about this scenario when we're showing conflict vs. synthetic tree. For mapped tips, we build a link to the synth-tree viewer on the assumption (it seems to me) that the taxon will be found there. This means once in awhile the user will get an empty view with explanation instead. Are we OK with that?

jimallman:
> I agree with @jar398 about including node IDs, and with removing redundancy (name multiple times in popup). I think simply saying "Mapped to" instead of "Aligned with" would remove confusion.

Agreed on all counts, though I'm hesitant to make the node id the most prominent (black, bold) element since we've tried to downplay these in the past.

More importantly, I'm trying to come up with a uniform presentation that handles all of the ways that nodes, labels, and taxa can be related:

screen shot 2016-03-10 at 1 54 55 pm

screen shot 2016-03-10 at 1 54 17 pm

screen shot 2016-03-10 at 1 53 31 pm

Still noodling...

Issue #902, opened 2016-02-18
Creating collections does not trigger creation of profile page?


kcranston:
For example, 
https://tree.opentreeoflife.org/curator/profile/bredelings

and
https://github.com/OpenTreeOfLife/collections-1/commit/c1456b801b048a58438ff55184b786bb4af1c175

@bredelings has no other activity (curation or feedback issues).

bredelings:
Also, currently https://tree.opentreeoflife.org/curator/profile/ hangs for me.

kcranston:
I wonder if we should add a "do you want to create a profile page?" option when a logged-in user is trying access their own profile page.

jimallman:
Thanks @bredelings, this would have bitten us eventually. We need to detect and recognize collection activity as project activity in [this general test for an OpenTree profile page](https://github.com/OpenTreeOfLife/opentree/blob/7a6c19386b582572a119087e048e99d463697c4f/curator/controllers/default.py#L197).

jimallman:
Yep, here's [my little reminder to add this when collections were done](https://github.com/OpenTreeOfLife/opentree/blob/7a6c19386b582572a119087e048e99d463697c4f/curator/controllers/default.py#L290). :bowtie:

jimallman:
> I wonder if we should add a "do you want to create a profile page?" option when a logged-in user is trying access their own profile page.

What's your intention here? To "force" the creation of a profile for someone who has no detectable activity? Or to let them keep a low profile (no profile page) regardless?

bredelings:
For me the main issue is discoverability of collections.  I kept on trying 'Manage my collections' (which would hang), but I guess if I had clicked 'Browse all collections', then that would have allowed me to make a new collection.  Currently 'Add new collections' doesn't do anything.

kcranston:
I propose that we remove the Tree collections menu and simply make it a link that opens the collections page. People can get to their collections from their profile page.

kcranston:
This is partially fixed with #905 (which removes the 'My Collections' menu item entirely). Leaving this issue open, because we do want collections activity to trigger profile page creation.

jimallman:
Pull request #913 fixes the original complaint and restores the **My Collections** menu.

Issue #874, opened 2016-02-02
Link to 'Add new collection' in collections menu broken


kcranston:
Says '#TODO'. Oops. ;)

PR #913, opened 2016-02-20
Collections in curator profile

jimallman:
This addresses a number of related bugs re: collections and the curator profile page:

- Adding or editing a tree collection is now recognized as OpenTree activity, so a user who has only done these things will have a proper curator profile page.

- The "their tree collections" list will appear for any valid OpenTree user.

- Fixed a server-side bug (500 error) when we show the "no such user in GitHub" page.

- We show tree-collection stats for an active curator, in a compact format.

- We show an explanation for missing study stats, pending [OTI fixes described elsewhere](https://github.com/OpenTreeOfLife/oti/issues/35).

- We restore the 'Tree collections' menu (in the app's main menubar) and enable 'Add new collection' from any page in the curation app.

### DIFF
--- a/curator/controllers/default.py
+++ b/curator/controllers/default.py
@@ -208,7 +208,8 @@ def _get_opentree_activity( userid=None, username=None ):
         'added_studies':[], 
         'curated_studies': [], 
         'curated_studies_in_synthesis': [], 
-        'collections':[]
+        'added_collections':[],
+        'curated_collections':[]
     }
     method_dict = get_opentree_services_method_urls(request)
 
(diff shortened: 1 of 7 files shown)


jimallman:
This fixes #902.

jimallman:
This also fixes #874.

kcranston:
Assuming that collections still suffer from the larger problem of not being able to associate activity with users that don't have a public email address on their GitHub profile?

jimallman:
Good question @kcranston! This activity is tracked through peyotl, so it likely behaves like study curation.

jimallman:
Two requested changes:
 - Remove "(details below)" if there are no collections to show
 - Remove link to pending oti issue, just a more general message here
   > Reliable counts are coming soon.

PR #925, opened 2016-02-22
Resolve monophyletic conflicting OTU mappings

jimallman:
~~**WORK IN PROGRESS**.~~ This feature is working now, but larger trees would benefit from a per-tree cache. (See commit messages for more details.) Addresses https://github.com/OpenTreeOfLife/feedback/issues/232

The current code is on **devtree** for review. As a demonstration, the example tree [provided in issue #232](https://github.com/OpenTreeOfLife/feedback/issues/232#issuecomment-186691172) has eight monophyletic sets of conflicting OTU mappings (e.g. _Gelasinospora tetrasperma_) that are detected and resolved automatically, i.e. an arbitrary node is chosen as the exemplar.

_**UPDATE:** For best performance, this now includes a per-tree cache of taxon-mapping info. This is flushed whenever we re-root a tree, mark/clear an exemplar node, or change any taxon mapping._

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -977,6 +977,7 @@ function loadSelectedStudy() {
             // support fast lookup of elements by ID, for largest trees
             viewModel.fastLookups = {
                 'NODES_BY_ID': null,
+                'TREES_BY_OTU_ID': null,
                 'OTUS_BY_ID': null,
                 'EDGES_BY_SOURCE_ID': null,
                 'EDGES_BY_TARGET_ID': null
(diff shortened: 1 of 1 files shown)


bredelings:
Maybe monophyletic conflicts should be not have an exemplar chosen, but just not be marked in red?

kcranston:
We didn't test this sufficiently and probably should not have been merged. If you open one of the trees with conflicting nodes from [example study in issue #232](https://devtree.opentreeoflife.org/curator/study/edit/tt_103?tab=trees) and look at the help text at the top of the screen, you see that the count is completely wrong, and we are highlighting nodes that do not have duplicate mappings. 

screen shot 2016-02-26 at 11 30 54 am

I _think_ what is happening (but I might be wrong) is that we are constructing a list of conflict in [showConflictingNodesInTreeViewer](https://github.com/OpenTreeOfLife/opentree/blob/eff8227f661e2ed05fdac849243c2af09acc4881/curator/static/js/study-editor.js#L3972), passing that to [showTreeViewer](https://github.com/OpenTreeOfLife/opentree/blob/eff8227f661e2ed05fdac849243c2af09acc4881/curator/static/js/study-editor.js#L3614), where we then [update the conflicts](https://github.com/OpenTreeOfLife/opentree/blob/eff8227f661e2ed05fdac849243c2af09acc4881/curator/static/js/study-editor.js#L3623) but don't update that `HIGHLIGHT_PLAYLIST` object. 

Anyway, this shouldn't be deployed until fixed. Pinging @jar398 and @jimallman

jimallman:
Ouch! I'll fix this ASAP. (Your diagnosis sounds dead-on. Not sure why these are all being included in the HIGHLIGHT_PLAYLIST, but it should be a quick fix.)

PR #936, opened 2016-02-27
Revert "Resolve monophyletic conflicting OTU mappings"

kcranston:
Reverts OpenTreeOfLife/opentree#925

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -977,7 +977,6 @@ function loadSelectedStudy() {
             // support fast lookup of elements by ID, for largest trees
             viewModel.fastLookups = {
                 'NODES_BY_ID': null,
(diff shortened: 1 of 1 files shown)


kcranston:
I suggest we simply revert this PR (I didn't know you could do this!) and then re-merge the feature branch later. This allows us to re-deploy while we figure out the OTU conflicts and caching. (It also lets me go ahead with another feature branch that touches some of the same code without creating too much mess). Ok with @jar398 and @jimallman ?

jimallman:
> Ok with @jar398 and @jimallman ?

Unless I'm missing something, any current concerns are addressed by moving forward with PR #939, which fixes the "false positives" (nodes wrongly marked as having conflicting OTU mapping) in the tree view.

PR #939, opened 2016-02-27
Fix for clumsy per-tree cache of conflicting OTUs

jimallman:
I was caching *all* OTU mapping info, instead of conflicting mappings only (as intended). This resulted in lots of false positives in the tree view, i.e. nodes marked as conflicting that were not.

This latest version (IMO) demonstrates that we don't need to revert a prior merge in #936.

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -7758,17 +7758,18 @@ function getConflictingNodesInTree( tree ) {
 
     if (!isPreferredTree(tree)) {
         // ignoring these for now...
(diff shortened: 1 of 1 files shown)


jar398:
I think the cache is super high priority - I've seen how much of a performance improvement it gives. (Hope it still does so after this fix.) Is it running on devtree? What should I do to review it?

jimallman:
> Is it running on devtree? What should I do to review it?

Yes, but technically that was from the previous branch. Let me merge+push this now, just for peace of mind...

PR #946, opened 2016-03-27
Support cross-domain request for Nexson conversion

jimallman:
Initial use case is to provide conversion (from Newick, etc) from the Tree Illustrator prototype.

This code has been running for some time on **devtree**, but was never properly merged to master. I noticed it was missing when conversion calls from Tree Illustrator started failing, and I realized it was because we had recently deployed a different branch.


jimallman:
Note that this has been merged into branch `adapt-to-new-arguson` (the branch that uses v3 APIs), to support simultaneous work on v3 and the Tree Illustrator on **devtree**.

PR #954, opened 2016-04-07
change download link in properties pane

kcranston:


kcranston:
pinging @jimallman about whether a simple relative path is the best way to display this URL

jimallman:
> whether a simple relative path is the best way to display this URL

Versus root-relative, you mean? This might be tricky, since a synth-tree-viewer URL has optional path parts. I'll see if it always redirects to a consistent format...

jimallman:
Good and bad news. Happily, any of these partial URLs will be redirected to this consistent URL:
`https://tree.opentreeoflife.org/opentree/argus/opentree5.0@ott93302`
I've tested all these:
- https://tree.opentreeoflife.org/
- https://tree.opentreeoflife.org/opentree/opentree5.0@ott93302
- https://tree.opentreeoflife.org/opentree5.0@ott93302
- https://tree.opentreeoflife.org/argus/opentree5.0@ott93302

BUT, a URL generated from TNRS (taxon search) does not redirect, and it's different enough that it breaks the pattern:
`https://tree.opentreeoflife.org/opentree/argus/ottol@93302/cellular-organisms`
... so in short, we should use a root-relative path here.

Issue #871, opened 2016-02-02
link to github from collection  


kcranston:
Should be to item in github repo, raw, e.g. [regular github link](https://github.com/OpenTreeOfLife/collections-1/blob/master/collections-by-owner/kcranston/barnacles.json) and [raw](https://raw.githubusercontent.com/OpenTreeOfLife/collections-1/master/collections-by-owner/kcranston/barnacles.json)

kcranston:
I think this might actually be a peyotl issue, but I am having a hard time following the code. The relevant collection property is [`external_url`](https://github.com/OpenTreeOfLife/opentree/blob/17f429a8f5dc42ba37c7cee7f3f1ce09aa219f73/curator/views/layout.html#L557). This comes from [`get_external_url`](https://github.com/OpenTreeOfLife/peyotl/blob/7086a4a2b7ea242877038366cae5497a00ce7fd7/peyotl/api/collections_api.py#L76) in peyotl, set in [`_remote_external_url`](https://github.com/OpenTreeOfLife/peyotl/blob/7086a4a2b7ea242877038366cae5497a00ce7fd7/peyotl/api/collections_api.py#L192). This URL gets used  in github interactions, so perhaps the solution is a new property for collections? `view_url` or something similar?

jimallman:
Thanks, I'll take a look!

jimallman:
>  perhaps the solution is a new property for collections? view_url or something similar?

Agreed, maybe something like `repo_url` or `source_repository_url`? I could see this being useful for all types in peyotl, actually (alongside `external_url` wherever it appears).

jimallman:
If that's too much, we could munge `external_url` to get to its main page in the repo. Just change this:
`https://raw.githubusercontent.com/OpenTreeOfLife/collections-0/master/`
...to this:
`https://github.com/OpenTreeOfLife/collections-0/blob/master/`

kcranston:
I was checking to see how we handle this with studies, but we don't actually provide a link to the study in github (other than the diff view in the history tab).

jimallman:
See the comments on #964 for alternative solutions, esp. changing the URL we provide, or providing separate "download" and "history" URLs.

PR #964, opened 2016-04-17
Reconstruct full GitHub URL for collections

jimallman:
We can safely convert between a download URL on GitHub (currently provided as the collection's `external_url`) and a "full" URL to the collection blob and its history. Addresses #871.

**N.B.** that we might instead choose to [redefine all "public" or "external" URLs](https://github.com/OpenTreeOfLife/peyotl/blob/a91833404b6e84ac33ebbc176296bdf9a2c698fc/peyotl/git_storage/sharded_doc_store.py#L34-L39) in peyotl's documents, if we'd **always** prefer to show the richer GitHub view. Or we might distinguish between the two, if we can agree on suitable property names, perhaps `doc.download_url` (looks like [this](https://raw.githubusercontent.com/OpenTreeOfLife/collections-0/master/collections-by-owner/jimallman/new-test-ad.json)) versus `doc.history_url` (looks like [this](https://github.com/OpenTreeOfLife/collections-0/blob/master/collections-by-owner/jimallman/new-test-ad.json)).

### DIFF
--- a/curator/static/js/curation-helpers.js
+++ b/curator/static/js/curation-helpers.js
@@ -706,7 +706,7 @@ $(window).resize( function () {
 });
 
 // Use a known-good URL fragment to extract a collection ID from its API URL
-var collectionURLSplitterAPI = '/v2/collection/';
+var collectionURLSplitterAPI = '/collection/';
 // Fall back to raw-data URL in some cases
 var collectionURLSplitterRaw = '/collections/';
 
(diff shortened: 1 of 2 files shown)


jimallman:
The simple fix above (munging the download URL) is working now on **devtree**.

kcranston:
This is fine to me as a temporary solution. I've made a [separate issue in peyotl](https://github.com/OpenTreeOfLife/peyotl/issues/149) for the question of having explicit view URLs for studies and collections.

Issue #883, opened 2016-02-04
Better links to user profile pages


kcranston:
Two things:

1. There isn't any obvious way to get to your user profile page. The only link is the 'Manage my collections' item in the Collections menu while in the curator app. It would be good to have an explicit My Profile link accessible from both webapp and curator. 
2. There isn't any way to get to someone else's profile page without constructing the URL yourself
3.  The URL does not change when you are on your profile page - it is always the generic `https://tree.opentreeoflife.org/curator/profile`. The more specific link to `https://tree.opentreeoflife.org/curator/profile/username` works, but there isn't any way to navigate to that page via links in the app.

kcranston:
Ok, just found the profile link under my name while logged into the curator app. So, 1 above isn't true - there is an explicit link. Second point still stands, though.

jimallman:
> There isn't any way to get to someone else's profile page without constructing the URL yourself

There's no explicit "user search", if that's what you mean. We link curator names to these profile pages in the collections UI, and we can certainly do so from other places. (I believe the main study list uses these curator names to filter the list by curator, but these could instead link to each curator's profile page.)

Re (3) above: We can redirect the my-profile URL (`/curator/profile`) to the more explicit and sharable URL, if you like.

jimallman:
> We can redirect the my-profile URL (/curator/profile) to the more explicit and sharable URL

This is working now on **devtree**, which should make it easier to share your own profile.

We should talk about what to offer in terms of user search.. Perhaps this could be added to ottreeindex? It will need to build the userid list by calling the GitHub API, or perhaps by gathering contributors from all of our feedback issues and phylesystem documents.

kcranston:
Couple of thoughts (some broader than this specific issue, so probably needs a new issue):

1. User search lower priority that many other open issues
1. Having a searchable user index will happen as a side effect of providing the inverse query (providing a list of curated studies on a user profile page, rather than the "Reliable counts are coming soon" message)
1. We probably want to more strongly encourage people to set up their github profiles to allow github to correctly link their curation activity to their github accounts

I think we can safely close this issue, though. Shareable URL success!

PR #965, opened 2016-04-18
Always redirect to user's (default) profile page...

jimallman:
...if the userid wasn't specified on the inbound URL. (We use the simple link in menus, help text, etc., while the expanded URL is better for sharing.) Addresses #883.

This is working now on **devtree**, at https://devtree.opentreeoflife.org/curator/profile


kcranston:
Looks good to me. Thanks!

Issue #920, opened 2016-02-21
Curators without public email addresses


kcranston:
Couple of requests here:

- [ ] add a warning message the first time a user without a public email address tries to save a study. Something like "If you don't have a public email address on your GitHub profile, GitHub will not associate your curation activity with your GitHub account"
- [ ] add a information message to the curator page "why do I not have any curated studies even though I have curated studies?" (in the short term, until we do the next one)
- [ ] scrape the phylesystem log and match usernames so that we can provide a list of curated studies even in cases where the user has no github activity that we can associate with their account

josephwb:
FWIW I have purposely not added a public email to my github profile so that you can use my account to troubleshoot things.

PR #967, opened 2016-04-20
Warn curator without public email addresses

kcranston:
This addresses part of #920, providing a warning message on the commit popup when a user does not have a public email address on GitHub. Deployed on devtree, but a PITA to test (you need to make your email address private on GitHub, re-load devtree (maybe clearing cache?), log out and log in again, make a change to a collection or study and save). You can confirm the status of your email address on your opentree profile page - public GitHub email = no email listed on profile page. For anyone not thrilled about going through all of that hassle, this is what the new popup looks like when you don't have a public email address:

screen shot 2016-04-20 at 11 05 56 am


kcranston:
I clearly messed up git with this one. Let me try and fix and then make a new PR.

PR #968, opened 2016-04-20
Warn user no email

kcranston:
This addresses part of #920 , providing a warning message on the commit popup when saving studies or collections and when a user does not have a public email address on GitHub. Deployed on devtree, but a PITA to test (you need to make your email address private on GitHub, re-load devtree (maybe clearing cache?), log out and log in again, make a change to a collection or study and save). You can confirm the status of your email address on your opentree profile page - public GitHub email = no email listed on profile page. For anyone not thrilled about going through all of that hassle, this is what the new popup looks like when you don't have a public email address:

screen shot 2016-04-20 at 11 05 56 am

### DIFF
--- a/curator/static/js/curation-helpers.js
+++ b/curator/static/js/curation-helpers.js
@@ -1334,7 +1334,14 @@ function promptForSaveCollectionComments( collection ) {
         $('#save-collection-comment-first-line').val(firstLine);
         $('#save-collection-comment-more-lines').val(moreLines);
 
-        $('#save-collection-comments-popup').modal('show');
+        if (userEmail != 'ANONYMOUS') {
+          // console.log('email: '+ userEmail);
+          $('#save-collection-comments-popup').modal('show');
+        }
+        else {
+          $('#save-collection-comments-popup-no-email').modal('show');
+        }
+
         // buttons there do the remaining work
         $('#save-collection-comments-submit')
             .unbind('click')
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -2225,7 +2225,14 @@ function validateFormData() {
 
 function promptForSaveComments() {
     // show a modal popup to gather comments (or cancel)
-    $('#save-comments-popup').modal('show');
+    // check for valid user email address and show modal with warning if none
+    if (userEmail != 'ANONYMOUS') {
+      console.log('email: '+ userEmail);
+      $('#save-comments-popup').modal('show');
+    }
+    else {
+      $('#save-comments-popup-no-email').modal('show');
+    }
     // buttons there do the remaining work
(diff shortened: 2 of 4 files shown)


curator/static/js/curation-helpers.js
```diff
@@ -1334,7 +1334,14 @@ function promptForSaveCollectionComments( collection ) {
...
         $('#save-collection-comment-more-lines').val(moreLines);
 
-        $('#save-collection-comments-popup').modal('show');
+        if (userEmail != 'ANONYMOUS') {
+          // console.log('email: '+ userEmail);
+          $('#save-collection-comments-popup').modal('show');
+        }
+        else {
+          $('#save-collection-comments-popup-no-email').modal('show');
+        }
```
jimallman: Due to the presence of (presumably) unique IDs within this popup, it's probably safer to always use the original popup element and hide/show a warning element inside it. Otherwise on submission we're liable to lose the comment text in one case or the other.

kcranston:
Three comments here:
1. In response to @jimallman comment - this was the route that I started to take, but wasn't sure how, so this was option 2
1. The save & comment seems to work both with and without email address: [commit with](https://github.com/OpenTreeOfLife/phylesystem-0/commit/9de582eb06df51683d282e7fd9f367f1165ffff4) and [commit without](https://github.com/OpenTreeOfLife/phylesystem-0/commit/924943b9c7932bc2f01c6cb61ca0189a72ae80c7)
1. There is an outstanding issue: the modal closes after Save when (public email, no message displayed) but does not close on Save when (private email, message displayed). Need to investigate why, but suggestions welcome.

jimallman:
Apologies for my delayed response! I'm not surprised there were problems with this change, since we depend heavily on ID-based selectors in jQuery. These generally assume that there's just one element in the current document with the given ID, and in this case we often have two (e.g., `#save-collection-comment-first-line`): one in the active window, and another in the inactive HTML template. Avoiding these bugs would mean assigning distinct IDs in _all_ cases, checking our event handlers, etc.

I'd recommend the simpler solution mentioned above:
> use the original popup element and hide/show a warning element inside it.

This means we'd make these changes:
  - move the warning element (`div.help-box` and the surrounding `p`) into the original popup template
  - remove the new template `#save-collection-comments-popup-no-email`
  - give the `p` around the new warning a unique ID, perhaps `no-email-warning`?
  - simplify the new conditional logic to always use the original popup, but to show (or hide) the warning based on the absence (or presence) of a public email:
```javascript
        if (userEmail != 'ANONYMOUS') {
          $('#no-email-warning').show( );
        } else {
          $('#no-email-warning').hide( );
        }
```
If this approach sounds acceptable, I'll be glad to make these changes.

Issue #881, opened 2016-02-04
Collections should have URLs


jar398:
When I look at a collection, the browser URL bar says 'https://tree.opentreeoflife.org/curator/collections'. There needs to be a URL specific to each collection so that people can bookmark them, refer to them in issue reports, and so on.

jimallman:
I have a simple solution working now on **devtree**. This includes minor revisions to the collection UI, including a "Share" button:

![screen shot 2016-04-15 at 4 44 23 pm](https://cloud.githubusercontent.com/assets/446375/14574399/7803a2b4-0329-11e6-981a-f63bdc101d9b.png)

This provides a direct URL to the current collection, something like this:
https://devtree.opentreeoflife.org/curator/collections/jimallman/my-test-collection
(This opens the main collections page and auto-loads the specified collection.)

Note that I'm not otherwise modifying the browser's URL or manipulating browser history as we do when viewing trees, etc. This is because the collection popup can appears on several different pages, most of which have other history-manipulation features tied to tabs, list filters, etc. It seems logical that the "canonical" URL for a collection should be in the form shown above. 

As always, thoughts and questions are welcome.

kcranston:
I am tempted to ask for either the collection title or collection ID to be a link to this URL, but I suspect this might create history problems. Am I correct?

jimallman:
> I suspect this might create history problems. Am I correct?

History will be handled sensibly in both windows, but there's a real danger (likely enough to give me pause) that someone will edit the same collection in multiple windows, with conflicting edits and a mess to clean up later.

jimallman:
But if you mean that someone will open the collections direct URL in **the same window**, that would work fine. It would just take us out of the original context (e.g., the study editor or profile page) with possible confusion for the user.

jimallman:
I've tried to address the suggestion above with this revised UI (deployed to **devtree**):

<img width="586" alt="screen shot 2016-04-29 at 12 18 28 pm" src="https://cloud.githubusercontent.com/assets/446375/14922429/8a4b5264-0e05-11e6-84cc-d06b6a2ae3ea.png">

The new History button shows the commit history (if any) for this collection, in a new window. As requested, the collection ID also now provides a direct link to the collection in the curation app. This makes the existing Share button redundant, but maybe it's a good idea to make this easy to find..?

[...]

PR #973, opened 2016-04-29
Misc collection fixes (direct and history URLs)

jimallman:
Add support for a "direct URL" for each collection, and another to its history in GitHub. Revise collection popup UI to show these URLs. See commit messages for details. This is all working now on **devtree**.


jimallman:
See related discussion in #881 and #964.

PR #990, opened 2016-06-05
update to v3 apis; add search by source id

jar398:
1. The browse script now uses the v3 taxonomy and TNRS APIs. 400 responses are handled in a halfway plausible way.
2. You can put a source id (e.g. ncbi:1) in the search box.


jar398:
This is running on devtree in case you'd like to try it out: https://devtree.opentreeoflife.org/taxonomy/browse

PR #997, opened 2016-07-04
add extinct as feedback option per https://github.com/OpenTreeOfLife/…

jar398:
For motivation see https://github.com/OpenTreeOfLife/feedback/issues/59.  We currently have 4 open and 8 closed feedback issues simply noting that a taxon is extinct, and we should encourage people to submit lots more of them. Feel free to wordsmith and so on.

### DIFF
--- a/webapp/controllers/plugin_localcomments.py
+++ b/webapp/controllers/plugin_localcomments.py
@@ -72,6 +72,7 @@ function capture_form() {
                 break;
             case 'Correction to relationships in the synthetic tree':
             case 'Correction to names (taxonomy)':
+            case 'Extinct/extant issue (taxonomy)':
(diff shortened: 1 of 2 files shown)


jimallman:
@jar398 I just noticed this PR. The feedback moderation page is currently unused, but I'll add the equivalent to the dropdown menu, possibly in time for today's PR review. :clock2:

jar398:
Oops! O guess I put it in the wrong place!  Thanks for looking at it.​  I
thought I had done a search that would find the right code but I guess not.

jimallman:
The actual feedback UI is constructed using web2py's DOM-building methods, so it defies easy searching. I've added the new type to its controller (`plugin_localcomments.py`), plus some logic to make sure that the supporting-evidence field appears when it's chosen. This works in my local testing.

jimallman:
This is working now on **devtree** as well.

PR #1016, opened 2016-08-11
Fix broken links to trees (References page)

### LINKED ISSUE #1011: Malformed links to single tree
See links on right-hand side: https://tree.opentreeoflife.org/about/references

These are looking for IDs like 'treetree123', where the actual tree ID is 'tree123'.

jimallman:
Fixes #1011, by removing the standard 'tree' prefix from tree IDs when linkin from [the References page](http://devtree.opentreeoflife.org/about/references). This works properly now, which you can confirm by testing against these studies near the top of the list:
  - ~~Quintero et. al. (uses `tree*` for tree ids)~~
  - ~~da Marcela Vasco-Palacios et. al. (uses `Tr*` for tree ids)~~

~~**Note** that most links from the References page on **devtree** will fail, as they point to the dev phylesystem. To test these properly, adjust the web address of the _destination page_ by changing 'devtree' to 'tree'.~~

EDIT: The examples and cautions above were due to quirks in my local test system. Our shared **dev** system is internally consistent, so links from the References page should behave normally. I don't see any studies that use non-standard tree ids, but my local testing shows that they would be supported.

### DIFF
--- a/webapp/views/about/references.html
+++ b/webapp/views/about/references.html
@@ -92,8 +92,8 @@ div.contributing-studies div.links .tree-links a {
                       ><span>Show links to included trees</span> ({{= len(study['tree_ids']) }})</a>
                   <div class="tree-links">
                      {{ for tree_id in study['tree_ids']: }}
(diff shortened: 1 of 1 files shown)


jimallman:
This is deployed on **devtree** and should work as expected.

PR #1017, opened 2016-08-12
Link taxonomic amendments to source docs on GitHub

jimallman:
We link to all other taxon sources in the taxonomy browser, so I thought we should try to clarify those added by taxonomic amendments. Here we link directly to the latest version of the amendment JSON on GitHub (after sniffing the API base URL to choose the appropriate source repo). 

This is obviously not pretty, but it will quickly answer particular questions about why a taxon was added, and by whom.

This is working now on **devtree**, see for example [this test entry 'Felis polydacta'](https://devtree.opentreeoflife.org/taxonomy/browse?name=Felis%20polydacta).

### DIFF
--- a/taxonomy/cgi-bin/browse.py
+++ b/taxonomy/cgi-bin/browse.py
@@ -2,11 +2,11 @@
 
 # Brutally primitive reference taxonomy browser.
 # Basically just a simple shim on the taxomachine 'taxon' method.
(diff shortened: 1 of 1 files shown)


jimallman:
As discussed, I've simplified the link text for amendments to the form 'addition:1234567', and added whitespace before the source list for each taxon.

Issue #969, opened 2016-04-25
should have link from taxonomy release page to taxonomy browser


kcranston:
https://tree.opentreeoflife.org/about/taxonomy-version/ott2.9

jimallman:
This would only work for the _current_ taxonomy, right? Is your intention that we'd try to support browsing _all_ versions of the taxonomy?

kcranston:
I just wanted a link to the current version.

jimallman:
Just to be clear: So this link would appear on the page describing the current/latest version, and would be omitted on previous versions. If this is correct, that should be an easy fix.

kcranston:
I suppose so. They way that the page is set up (for example, the title is "Taxonomy version", not "Open Tree Taxonomy"), there isn't an obvious place to put a link to the current version that wouldn't seem strange when you weren't looking at the current version.

jimallman:
Please see PR https://github.com/OpenTreeOfLife/opentree/pull/1018 . I'm hoping the solution shown there will work for both new and old OTT versions.

PR #1018, opened 2016-08-12
Link OTT version to taxo browser

jimallman:
Addresses #969. This uses a convention I've seen on other sites, where the current/latest version shows a "success" notice (green), and older/deprecated versions show a yellow- or red-tinted notice.

screen shot 2016-08-12 at 2 02 15 am

screen shot 2016-08-12 at 2 02 32 am

This is working now on **devtree**.


kcranston:
I like it! Thanks!

PR #1032, opened 2016-08-18
Davids annoyances

kcranston:
Minor changes based on feedback from David Maddison:

1. Better wording on collections edit screen
1. Remove unused fields on study metadata tab
1. Fix URL for web APIs

### DIFF
--- a/curator/views/study/edit.html
+++ b/curator/views/study/edit.html
@@ -612,8 +612,7 @@ body {
 
                 <p>
                 You can also use the <strong>
-                <a href="https://github.com/OpenTreeOfLife/opentree/wiki/Open-Tree-of-Life-APIs" target="_blank">Open Tree API</a>
-                </strong> (the same one that powers this editor) to retrieve and make changes to study data.
+                <a href="https://github.com/OpenTreeOfLife/germinator/wiki/Open-Tree-of-Life-Web-APIs" target="_blank">Open Tree API</a></strong> to retrieve study data.
                 </p>
 
             </div>
--- a/curator/views/study/edit.html
+++ b/curator/views/study/edit.html
@@ -513,7 +513,7 @@ body {
                 </div>
               </div>
 
-              <div class="control-group">
+              <!-- <div class="control-group">
                 <div class="controls">
 
                 {{ if viewOrEdit == 'EDIT': }}
@@ -529,7 +529,7 @@ body {
 
                 </div>
 
-              </div>
+              </div> -->
 
               <div class="control-group">
                 <label class="control-label">Curator notes</label>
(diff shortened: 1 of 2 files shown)


curator/views/study/edit.html
```diff
@@ -513,7 +513,7 @@
                 </div>
               </div>
 
-              <div class="control-group">
+              <!-- <div class="control-group">
```
jimallman: Just a reminder that the `^ot:notIntendedForSynthesis` flag does have a minor effect in the curation app. If it's checked, we relax the "Study quality" scoring by skipping (passing) these tests:
  - For studies nominated for synthesis, there should be at least one preferred tree.
  - Preferred trees should not have duplicate (non-monophyletic) tips mapped to a single taxon.
  - All tip labels in preferred trees should be mapped to the Open Tree Taxonomy.

Without this checkbox, all studies will be subject to more stringent tests by default.

jimallman:
As requested, I've modified some button labels, particularly the 'add tree' button which now shows one of two labels:
  - if a collection is empty: **Add a tree to this collection**
  - if there's at least one tree already: **Add another tree**

Note that removing the last remaining tree will correctly update the button text. This is available for review on **devtree**.

jar398:
I have really been confused about how the candidate-for-synthesis study and tree flags worked and am glad this PR made me review them.

Given the meaning of the candidate-for-synthesis tree flag (as implied both by its Nexson name and the help text on the right) I wonder if we could change the column heading. Currently it is "Preferred" which is less than helpful (preferred to what? for what purpose?). It should be any phrase that is consistent with the help text. It could be as short as "Synthesis" or "Candidate" - I know that the table is starved for horizontal space.

I think there ought to be a way for a curator to distinguish "the trees have been reviewed and none is a synthesis candidate" from "the trees have not been vetted for inclusion in synthesis". It is this distinction that the notIntendedForSynthesis flag (removed by this PR) made in my mind, so if notIntendedForSynthesis is to go away I would like to see it replaced by some other way to express this meaning. If the distinction isn't communicated to a later curator, the later curator risks either overlooking good trees (if they weren't vetted) or redundantly doing the work of vetting bad trees (a waste of curator effort).

I slightly prefer keeping notIntendedForSynthesis but moving the control to the trees-list page (so it's visible at the same time as the per-tree checkboxes), and labeling it something like "Do not nominate any tree for synthesis" or "No tree in this study should contribute to synthesis". If any tree is checked, the no-tree checkbox would get cleared. Another possibility is to make the per-tree flag be three-way: candidate, not candidate, and not yet vetted.

jar398:
Discussing of tree selection moved to https://github.com/OpenTreeOfLife/opentree/issues/1042

Issue #1034, opened 2016-08-18
Confusion about edited label in New Taxon window


kcranston:
When you select a taxon and open the New Taxon window, the `New taxon name` field is populated with `Modified for mapping` label from the OTU Mapping tab (or with the `Original taxon label`, if no modified label exists). If the user then deletes that text, then are stuck in a strange state of not being able to say "I don't want to modify the label". Opening and closing the popup doesn't work, because the state is saved. I wonder if "I want to submit a modified label" should be an explicit option to avoid this confusion. 

In addition, can we just use red asterisks to indicate required fields, rather than the red boxes, at least before a user tries to hit submit? Otherwise, it looks like the form is full of errors even if filled out correctly. 

<img width="564" alt="screen shot 2016-08-18 at 2 56 53 pm" src="https://cloud.githubusercontent.com/assets/312034/17786708/1b4c761c-6554-11e6-9926-2742df8a7093.png">

jimallman:
> If the user then deletes that text, then are stuck in a strange state of not being able to say "I don't want to modify the label".

What is the user's intent in this case? If they'rey trying to use the original label, I had assumed they would copy + paste from previous line. We could add a button to do this, I suppose, and enable it whenever the "New taxon name" doesn't match the original label. Or we can (implied in your answer) use the original label if the "New taxon name" field is empty, but this seemed pushy to me.

jimallman:
See updates and screen grabs in PR #1036

PR #1036, opened 2016-08-22
Better feedback for new-taxon naming

jimallman:
Addresses #1034. Includes these changes:
 - track label changes versus **original** label
 - add button to use/restore original label
 - smarter validation logic for label changes
 - remove unsightly red border on required fields

### DIFF
--- a/curator/static/css/default.css
+++ b/curator/static/css/default.css
@@ -865,7 +865,7 @@ tr.after-shims th {
 }
 
 .required-field {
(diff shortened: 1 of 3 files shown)


jimallman:
Here's a grab of the new, explicit button that restores the original label:

screen shot 2016-08-22 at 7 51 22 pm

jimallman:
This is now working on **devtree**.

Issue #1037, opened 2016-08-23
Curator regex doesn't fully handle remove last word 


josephwb:
While mapping tips with a trailing extra ID I noticed that the premade regex for "remove last of multiple words" doesn't work if there is a "-" in the word:
![regex](https://cloud.githubusercontent.com/assets/808499/17897686/1444e656-6923-11e6-8c4b-0bf5f7f94db4.png)
So the regex just needs to add "-" to the pattern. Change:
```
\b\s+(\w*)$
```
to:
```
\b\s+([\w-]*)$
```
I am happy to do this, but do not know where the relevant code lives.

Looks like this would need to be added to "remove first of multiple words" as well.

jimallman:
Yes, the built-in substitutions are [well hidden in the edit page](https://github.com/OpenTreeOfLife/opentree/blob/51beecf7b91636c2e4e8abc455653c9b7e0c4832/curator/views/study/edit.html#L1600-L1605).

Capturing the dash should be an easy change in several of these, but I wonder if we'd be better off specifying all non-whitespace characters. This would capture "words" like these:
  - 12345
  - 123-45
  - 123_45
  - 123:45

Do you think this is a better match for the likely "words" we'll see in these labels?

jimallman:
Reminder: These default substitutions are added to any incoming study that doesn't have them already. An existing study will have its own substitutions (stored in Nexson) so your changes won't be reflected.

jimallman:
If the rule above makes sense (match all non-whitespace characters), just change `\w` to `\S` in all cases, e.g.:
```
\b\s+(\w*)$
```
to:
```
\b\s+(\S*)$
```

josephwb:
I was thinking along the same lines (although I think `\w` includes the underscore). I just pushed a new branch with this edit. :bowtie:

josephwb:
I think this was merged today.

PR #1038, opened 2016-08-23
moar inclusive regex to match words

josephwb:
Built-in regex patterns for removing first/last words were failing on matching when a dash was involved in the string. These two changes will match any non-whitespace characters, so should work in all circumstances.  :pineapple:


jimallman:
I've also made the 'Isolate trailing IDs' substitution more inclusive. This is now running on **devtree** and looks good to me in some quick testing.

jimallman:
See related discussion at issue #1037 regarding what a likely "word" is in found tip labels.

josephwb:
Doh, I missed the "Isolate trailing ID". Thanks for getting that @jimallman!

Issue #381, opened 2014-07-22
Include common diacritical substitutions in client-side search


jimallman:
For example: Searching for "Lutzen" should find "Lützen". Since lots of this kind of filtering is done in client-side Javscript, we'll  to add a series of diacritical substitutions to search for wherever a “questionable” letter is found. The result is probably a regexp with all possible substitutions in place of each simple letter, for example:

Lutzen  ==>  /L[uü]tz[eé]n/

jimallman:
Naturally, we're not the first to need this. [Here's a pretty thorough fix](http://stackoverflow.com/a/228006) that probably does a bit too much splitting phrases.

Note that the solution above would preserve an accented character and not match its non-accented counterpart. It seems desirable for a search "Lützen" to also find "Lutzen".

[Here's another attempt](http://stackoverflow.com/a/5913376) that strips/reduce accented characters; perhaps we could adapt both into a two-pass system that

 - detects any variant (accented or not) and substitutes the unaccented version (or a token)
 - detects unaccented chars (or corresponding tokens) and substitutes with the search-all [...]

jar398:
FWIW, accents and diacritics are removed from taxon names before they go
into the taxonomy (they are disallowed by the nomenclatural codes). They
only occur in synonyms.

jimallman:
See also this related article [_Accent Folding for Auto-Complete_](http://alistapart.com/article/accent-folding-for-auto-complete) and the more exhaustive [RFC 3454 - Preparation of Internationalized Strings](https://tools.ietf.org/html/rfc3454).

jimallman:
I settled on an approach that applies all common diacritical "variants" when searching, so a search entered as `aether` will actually use this regular expression:
```
[...]

PR #1047, opened 2016-09-10
Add diacritical variants for JS search + list filters

jimallman:
This solution will find diacritics (accented characters like `ñ`, `â`, `œ`) in the data, but missing from the query text, **and vice versa** (ie, cases where the original data was entered without the proper diacritical marks, but the searcher tries to use them). 

Since list filters also change the browser's history and the current URL, we also have a function `removeDiacritics` that "Latinizes" all values in the query string, to avoid encoding issues that cause the shared URL to fail. Since the searches now check for any diacritical marks, this value will return identical results to the original. 

I believe this handles most common diacritical variants; new entries can be added easily, using the same format as the entries in `diacriticalVariants` below.

This is deployed and working on **devtree**. To see how it works, try the following searches in [the main study list](https://devtree.opentreeoflife.org/curator/) of the curation webapp:

  - "fjeldsa" matches "Fjeldså" in a few studies
  - "jonsson" matches "Jønsson" in a few studies
  - "bøwie" (a nonsense spelling) matches studies with "Bowie"

In each case, note that the browser's address bar shows a "Latinized" version of the search string. If this URL is copied and pasted, it will show this latin-only string in the search/filter UI, but the filter/search results should be identical.

### DIFF
--- a/curator/static/js/curation-dashboard.js
+++ b/curator/static/js/curation-dashboard.js
@@ -252,7 +252,7 @@ function loadStudyList() {
                 var match = viewModel.listFilters.STUDIES.match(),
                     matchWithDiacriticals = addDiacriticalVariants(match),
                     matchPattern = new RegExp( $.trim(matchWithDiacriticals), 'i' ),
-                    wholeWordMatchPattern = new RegExp( '\\b'+ $.trim(match) +'\\b', 'i' );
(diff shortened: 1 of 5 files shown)


jimallman:
Oh, this fixes issue #381.

Issue #986, opened 2016-06-01
"Show All" details option on fron page of curator


josephwb:
A "show all" option for study details on the front page of the curator would be nice, I think. For example, I'd (often) like to see all study detail at a glance, and clicking on each "show details" is tedious.

jimallman:
I've added this feature, working now on **devtree**. See the new filter above the study list:

<img width="859" alt="screen shot 2016-09-15 at 12 18 28 am" src="https://cloud.githubusercontent.com/assets/446375/18538251/0302b878-7ada-11e6-8048-795e3ed81464.png">

As with our other list filters, changing this will modify the current URL (in the browser's address bar). Sharing this URL will also share your choice of compact/full format. This also makes it easy to reload the page, or use the Back button, to stick with your preferred view.

josephwb:
Looks good to me!

I do not recall requesting this, but: what a great idea!

josephwb:
I think this is good. Closing.

PR #1049, opened 2016-09-15
Implement a toggle to show all study details.

jimallman:
This is done as a stateful view ("compact" or "full"), so it can be shared in a URL and easily reloaded automatically. Addresses #986.

This is working now on **devtree**. See the issue above for more details.

### DIFF
--- a/curator/static/js/curation-dashboard.js
+++ b/curator/static/js/curation-dashboard.js
@@ -532,8 +532,7 @@ function getPubLink(study) {
         urlNotFound = true;
     }
     if (urlNotFound) {
-        return '<span style="color: #999;">No link to this publication.</span>';
-        //return '<span style="color: #ccc;">[DOI not found]</span>';
+        return "";
     }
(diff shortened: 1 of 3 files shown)


kcranston:
Two requests:
1. Can we remove this line from the full display: `This study has not been tagged. `
2. Can we put the publication link immediately after the publication reference (without a link break) and simply not show if not present?

kcranston:
Also, when you flip the toggle from compact to full, it puts you back to the first page of results

jimallman:
> Also, when you flip the toggle from compact to full, it puts you back to the first page of results

This is now fixed on **devtree**. As a side effect, pagination movements are now tracked in the URL (and added to browser history, which might be annoying for some). This is only true for the study list; other filtered lists in the curation app are unchanged.

Hm, should we reset to page 1 when changing the sort order (e.g. oldest first to newest first)?

jimallman:
> Can we remove this line from the full display: This study has not been tagged.
> Can we put the publication link immediately after the publication reference (without a link break) and simply not show if not present?

These changes are available for review on **devtree**.

Issue #998, opened 2016-07-10
Ability to download conflict results as images


josephwb:
It would be nice to be able to download images of the results of the conflict analyses (say, [this](https://tree.opentreeoflife.org/curator/study/view/ot_412/?tab=trees&tree=Tr77157&conflict=synth)). At the moment it is not possible to even use a screen grab, as most trees are so large that they involve scrolling to see the entire thing:

![untitled](https://cloud.githubusercontent.com/assets/808499/16710879/2938333e-460e-11e6-83aa-c21f03640259.png)

For example, I can see people wanting to include such an image in a talk or grant application etc.

jimallman:
See **devtree** and the PR above for new save and print features.

And late in the process I discovered [SVG Crowbar](http://nytimes.github.io/svg-crowbar/), a JS bookmarklet that can capture SVG from any web page, but only in Chrome. 😜 Neat!

jimallman:
Also, I'm hoping SVG is a friendly enough format for the downloaded image. Raster conversion to e.g. PNG is possible but would introduce lots of new moving parts. Comments welcome.

josephwb:
I do not have anything standard on my mac that can easily open SVG. Oddly, when I do open the SVG file on mac/Linux all I see are black branches (i.e. no colour) on a transparent background. This is probably just me.

PNG seems more accessible, but SVG is better quality. I do not know [which format is better](http://graphicdesign.stackexchange.com/questions/249/which-format-is-better-for-saving-images-intended-for-websites-png-or-svg).

jar398:
I believe any web browser can open an SVG.​ From there you can resize and
then use screen grab to get a TIFF.

When I download the SVG, the file does not get a file extension, so the
usual Mac opener doesn't work. But if you edit the file name to add 'svg'
to the end, it opens in safari, with color.

If the curator app has any way to set the extension, that would be nice.
Maybe there is a way to set a media type, and the downloader will then set
the extension from that?

inkscape and illustrator are options if you want to edit. It is also
possible to edit 'manually' (i.e. as a text file in a text editor).

jar398:
I agree that screen shots are a terrible way to capture images. I think I
would try Inkscape for high res SVG to PNG conversion, but maybe there is
an option that is simpler or better. Adding PNG functionality to the
curator app should be a last resort.

jimallman:
[...]

PR #1052, opened 2016-09-16
Allow printing (single page) tree views, and saving as SVG

jimallman:
This started as an attempt to address #998, but I mis-remembered it and was halfway to a print feature before I realized that a downloadable image file was really the goal. As it turns out, there's a lot of overlap between the two, and each feature seems useful, so I've kept them both. These appear in the footer of the curation app's tree viewer:

![screen shot 2016-09-16 at 7 07 37 pm]
(https://cloud.githubusercontent.com/assets/446375/18603593/e886d606-7c40-11e6-96f4-0c75252498ee.png)

In each case, the saved (or printed) tree will reflect the current layout and any displayed conflict information. The saved SVG is self-contained and editable in Inkscape (presumably also Illustrator, but confirmation would be great).

This is working now on **devtree**.  Tested in Chrome, Firefox, and Safari. Currently the printed tree is sized to fit on a single vertical page -- not ideal for a large tree, but some browsers are pretty... limited in how they handle SVG printing. _< Firefox :eyes: >_

### DIFF
--- a/curator/static/css/default.css
+++ b/curator/static/css/default.css
@@ -708,6 +708,33 @@ tr:hover .row-controls-ghosted {
     /* try to make this offsetParent to all its contents */
     position: relative;
 }
+body.printing-tree-view * {
+    display: none;
+}
+body.printing-tree-view svg,
+body.printing-tree-view svg * {
+    display: block;
+    background-color: #fff !important;
+    position: absolute;
+    z-index: 1000000 !important;
+    top: 50px;
+    left: 0;
+    /* border: 1px red solid; */
+}
+body.printing-tree-view #tree-title,
+body.printing-tree-view #tree-title * {
+    display: block;
+    position: absolute;
+    z-index: 1000000 !important;
+    top: 0px;
+    left: 0;
+    background-color: #fff !important;
+    margin: 0;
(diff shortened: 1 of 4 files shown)


jimallman:
Note that this seems to introduce some new cropping problems (labels cut off) on-screen, esp in radial layout, see for example https://devtree.opentreeoflife.org/curator/study/view/tt_100/?tab=trees&tree=Tr80848

I'll clean this up and we can revisit.

jimallman:
Hm, I think the observed cropping problem existed before these changes. It seems our bounding-box calculations are slightly off where longer node labels appear. Here are some examples on production (try both radial and normal layouts):

* https://tree.opentreeoflife.org/curator/study/view/ot_732/?tab=trees&tree=tree12
* https://tree.opentreeoflife.org/curator/study/view/ot_658/?tab=trees&tree=tree2

So it's a related problem, but not apparently due to these changes. Regardless, I'll try to clean it up in this branch.

jimallman:
It seems the latest commit fixes on-screen cropping, but printing (from Chrome) is badly cropped. Needs a closer look...

jimallman:
Fixed with last commit. Online and available for review on **devtree**.

PR #1057, opened 2016-09-22
Move studies w/o year to bottom of default sort.

### LINKED ISSUE #1056: change default sort order
We hoped this wouldn't be the case, but [this PR](https://github.com/OpenTreeOfLife/opentree/pull/1048) does indeed sort the studies without metadata to the top of the list. Evident on production:

screen shot 2016-09-21 at 9 01 38 pm

Can we change the default sort order so that these do not appear at the top of the list?

jimallman:
Naturally, these will now appear if the alternate sort 'Oldest publication first' is chosen. Still, better to hide this stuff in the default sort. Fixes #1056.


jimallman:
This (tiny) fix is now working on **devtree**.

Issue #1065, opened 2016-09-26
Relative doi link (links to ot study rather than pub)


josephwb:
Example: the PeerJ doi for [this study](https://tree.opentreeoflife.org/curator/study/view/ot_787/?tab=metadata).

Possibly relevant: this is a treebase import.

jimallman:
I'm looking at [other studies entered by @ningwang83](https://tree.opentreeoflife.org/curator/?match=ningwang83&view=Full+view+(show+details)), and it’s an odd mix of simple DOIs and URLs, mostly the former. 

I suspect we're failing to convert simple DOIs on TreeBASE import, though I could have sworn this was fixed some time ago.

jimallman:
See the original report of this problem in https://github.com/OpenTreeOfLife/opentree/issues/816.

It looks like this issue was kicked over to the peyotl repo (https://github.com/OpenTreeOfLife/peyotl/issues/138), where it sits undone. @mtholder Do you have time to close this one out?

jar398:
I find 74 studies with these unnormalized (i.e. non-http:) DOIs. They are
all ot_ studies and all are treebase imports. Not all studies imported from
treebase have the unnormalized DOI form, e.g. ot_701 doesn't. Perhaps the
DOIs that are normalized were auto-normalized as a side effect of an
unrelated edit to the study metadata?

    study/ot_02/ot_602/ot_602.json:"@href": "10.1016/j.cub.2014.03.030"
    study/ot_02/ot_802/ot_802.json:"@href": "10.3389/fpls.2016.00697"
    study/ot_03/ot_303/ot_303.json:"@href": "10.1038/ncomms5473"
    study/ot_08/ot_808/ot_808.json:"@href": "10.1038/srep25367"
    study/ot_11/ot_811/ot_811.json:"@href": "10.5061/dryad.30k4v"
    study/ot_13/ot_713/ot_713.json:"@href": "10.1093/sysbio/syu022"
    study/ot_15/ot_215/ot_215.json:"@href": "10.3767/003158510X494596"
    study/ot_17/ot_217/ot_217.json:"@href": "10.3114/sim.2009.64.04"
    study/ot_18/ot_518/ot_518.json:"@href": "10.1600/036364415X689249"
    study/ot_19/ot_519/ot_519.json:"@href": "10.1073/pnas.1511468112"
    ...

jimallman:
[...]

PR #1066, opened 2016-09-27
Fix bare DOIs (for pub reference AND data deposit)

jimallman:
More careful handling of DOI-or-URL fields `ot:studyPublication` and `ot:dataDeposit`. When loading a study, we'll convert bare DOIs to their URL form, if possible. This fixes bugs in both the
read-only and editable pages. Addresses https://github.com/OpenTreeOfLife/peyotl/issues/138, but not a complete solution.

In the read-only view of curation app, we've been making hyperlinks from user-entered values that are not bare DOIs _or_ proper URLs. Instead, we should just show these as the raw (string) value. The same fix applies to the main study list; we'll show invalid URLs _and_ bare DOIs as simple strings, instead of as broken hyperlinks.

### DIFF
--- a/curator/static/js/curation-dashboard.js
+++ b/curator/static/js/curation-dashboard.js
@@ -520,6 +520,8 @@ function getFocalCladeLink(study) {
 
     return '<a href="#" onclick="filterByClade(\''+ cladeName +'\'); return false;"'+'>'+ cladeName +'</a'+'>';
 }
+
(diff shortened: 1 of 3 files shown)


jimallman:
This is working now on **devtree**.

jimallman:
I should mention, this more directly addresses #1065.

Issue #999, opened 2016-07-13
Curator should list number of studies on front page


josephwb:
Finding out how many studies exist is not straightforward; earlier today @kcranston had to resort to:

```
find . -name "*.json" | wc -l
```

on a local clone. Seems like this number should be easily visible.

jar398:
I think this is a great idea - great for PR and transparency. But not sure how to implement. Is otindex the best place to get this information?

kcranston:
The [about method](https://github.com/OpenTreeOfLife/otindex/wiki/Otindex-and-Oti-differences#about) provides this info.

jimallman:
Yes, this is trivial to implement. I'm sort of stuck on choosing _where_  and _how_ to show it:
 - without losing much screen real estate
 - other promotional verbiage wanted here? expand/replace the big "help" block?
 - how much fanfare? this page is currently all about functionality, vs. promotion

jar398:
put it in some of that whitespace to the right of the 'add new study'
button?  I think a single simple understated declarative sentence (or
sentence fragment) would be completely adequate, maybe just "4423 studies"

josephwb:
Yeah, that sounds like a good place (next to "Add new study").

Maybe list both 1) number of studies and 2) number of trees.

jimallman:
> The about method provides this info.

@kcranston Is this method available now? I can't seem to find a working URL on dev or production.

jimallman:
I've put up a working study count on **devtree**, obtained by counting the complete study list we already fetch for our filtered list. 

I agree that the tree count would be a nice addition, but we don't currently include tree information in the study list so this is sketched in (as a second line) but commented out for now. I'll tinker with loading the study list with `verbose: true`... Since we cache the study list, this shouldn't really slow things down.

PR #1073, opened 2016-10-07
Show study count on main curation dashboard.

jimallman:
NB This also has placeholders for a tree count on the second line, but I need to find the fastest way to capture this information. (I'm reckoning the study count using the complete list loaded via AJAX.)

Addresses #999 (see further discussion there).


jimallman:
This is working now (study count only) on **devtree**.

Issue #1062, opened 2016-09-23
Prepare for changing "witness" in conflict service to synth node label rather than OTT id


jar398:
Currently a conflict analysis against the synthetic tree yields either "witness": "1234"  where "1234" is an OTT id, or no "witness" field.  I would like to change this to "witness": "ott1234" or "witness": "mrcaott123ott456".

To avoid a 'flag day' I propose changing the curator app so that it accepts _either_ "1234" _or_ "ott1234" for synthetic tree OTT-equivalent nodes.  (Later it can understand the mrca labels as well, but that's not essential at this point.)  Once that change is made, updating the conflict service can be done at leisure.

Does this sound reasonable? The alternatives are (1) a flag day, or (2) a fully compatible change, e.g. adding a "witness_label" field ("witness": "1234, "witness_label": "ott1234") (there are other options too). I've thought about (2) and find it kind of ugly and ungeneral.

jar398:
For analyses against the taxonomy, the "witness" would stay the same ("1234").

jimallman:
For ideal URLs (with '/opentree7.0@ott123' instead of '/synth@ott123'), we would probably:
 - move [`fetch_current_synthetic_tree_ids`](https://github.com/OpenTreeOfLife/opentree/blob/45a75fd190892290be6ff0c84da57f3a635e0e77/webapp/controllers/default.py#L123) into `modules/opentreewebapputil`, so it's available in the curation app, or
 - copy it into the curation app's [`study.py` controller](https://github.com/OpenTreeOfLife/opentree/blob/master/curator/controllers/study.py); then
 - call this to retrieve the latest synth-tree version and pass it to JS in the study-editor page [as shown in the main webapp](https://github.com/OpenTreeOfLife/opentree/blob/45a75fd190892290be6ff0c84da57f3a635e0e77/webapp/controllers/default.py#L66)

That is, unless we want to return this synthesis release string ('opentree7.0' or whatever) in the response from the conflict service itself.

jar398:
This seems to work just fine, and it also helps show that https://github.com/OpenTreeOfLife/reference-taxonomy/pull/255 is working.

PR #1074, opened 2016-10-08
Handle new-style synth-node IDs in conflict.

jimallman:
Avoid a zero-day situation by handling old- _and_ new-style node ids returned from the conflict service for the synthetic tree. Addresses #1062.

Note that we'd ideally like to incorporate the latest synthesis release in the resulting URL, e.g. 'opentree7.0'. This could be done by including this string in the study-editor page, or by returning it as an additional property from the conflict service.


jar398:
This looks about right, thanks!

Issue #995, opened 2016-06-20
'Back' fails to go back to taxonomy browser from synthetic tree browser


jar398:
1. Browse a synthetic tree page
2. Click on an 'ott:' link to go to taxonomy browser
3. Click on 'View this taxon in the current synthetic tree'
4. Click 'back'
   --> 'Sorry, there was an error checking for taxon status.'

jimallman:
Thanks. Looks like the problem that we use a redirect to expand the full argus URL, from
https://tree.opentreeoflife.org/opentree/ottol@93302
to
https://tree.opentreeoflife.org/opentree/argus/ottol@93302
The simplest fix is to specify the full URL in the taxonomy browser; otherwise we can modify the URL expansion to work in a way that doesn't add to history.

jimallman:
Hm, it looks like even pasting the complete (expanded) argus URL in a fresh tab will redirect once. No error message, but it takes two clicks to return to the taxonomy browser. Another reason to revisit the general logic behind this redirect.

jimallman:
I've made a tiny PR that should fix the error message (but it will still require two clicks to return to the taxo-browser).

jimallman:
> I've made a tiny PR that should fix the error message (but it will still require two clicks to return to the taxo-browser).

The extra step in browser history should be fixed by the PR above! This is working now on **devtree**.

PR #1090, opened 2016-11-01
Nudge incoming synth-tree URLs to latest synthesis.

### LINKED ISSUE #1031: Versioning in web application URL
From @hyanwong on gitter:

'By the way, I think the versioning of the website is not working, as https://tree.opentreeoflife.org/opentree/argus/opentree5.0@ott361838 produces the same web page as for opentree6.1, whereas the synthetic newick for v5 which I have here is lacking all those Chlamydomonas species at the plant root.'

jimallman:
This should happen in the event that someone uses a URL with an old synthesis version, or any other value besides the latest version or 'ottol'. We should redirect and show a one-time message to the user. Fixes #1031; look there for more description and a screenshot.

This change is live on **devtree**.

### DIFF
--- a/webapp/controllers/default.py
+++ b/webapp/controllers/default.py
@@ -76,6 +76,8 @@ def index():
     if incomingDomSource not in ('ottol', latestSyntheticTreeVersion, ):
         treeview_dict['domSource'] = latestSyntheticTreeVersion
         treeview_dict['nudgingToLatestSyntheticTree'] = True
+        # mark this as a redirect to a different resource
+        response.status = 303
 
     # when all is said and done, do we have enough information to force the location?
     if treeview_dict['domSource'] and treeview_dict['nodeID']:
--- a/webapp/controllers/default.py
+++ b/webapp/controllers/default.py
@@ -31,6 +31,8 @@ def index():
     treeview_dict['nodeID'] = ''
     treeview_dict['nodeName'] = ''
     treeview_dict['viewport'] = ''
+    treeview_dict['nudgingToLatestSyntheticTree'] = False
+    treeview_dict['incomingDomSource'] = 'none'
 
     # add a flag to determine whether to force the viewer to this node (vs. using the
     # browser's stored state for this URL, or a default starting node)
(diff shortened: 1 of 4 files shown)


kcranston:
Needs a little more testing before merge. Getting error (on Chrome and FF) for the base devtree.opentreeoflife.org URL (and continuing to get it while browsing from there)

jimallman:
@kcranston, @jar398 The latest changes are live on **devtree**. I believe they address all the concerns from yesterday's PR review:
 - short/default URLs use response code 200
 - URLs with old or invalid opentree version use 303 response, instead of 200
 - in no case will we add an extra step to browser history (fixes #995!)

kcranston:
Closer! Looks good if I browse to https://devtree.opentreeoflife.org

But, if I use a URL with an old version number, e.g. https://devtree.opentreeoflife.org/opentree/opentree5.0@ott541951/Equus then I get the message (correctly) for that node, but when I click on another node, the URL correctly changes `s/opentree5.0/opentree7.0/` but the message remains. (Chrome and FF).

jimallman:
> ...when I click on another node, the URL correctly changes s/opentree5.0/opentree7.0/ but the message remains.

I forgot to mention, this is because we're still on the same page. (I expected you would close the alert with the little "X" button on the right.) Possible remedies:
 - make the close gadget more obvious
 - add a timeout to the message, so it fades out after a few seconds
 - hide the message whenever we choose a new node in the tree (this seems cleanest)

kcranston:
Ha! I never thought to close the message. I wonder if the colour should be something other than red - it isn't an error as much as a friendly warning. 

I agree that hiding the message when we choose a new node is cleanest.

jimallman:
> I wonder if the colour should be something other than red - it isn't an error as much as a friendly warning.

Hm, the easy colors (provided by bootstrap) are a pale yellow and a pale green. The yellow was my first choice, but it doesn't really "pop" enough visually. I'll try green, and add the auto-hide behavior when moving between nodes. Thanks for the quick feedback!

jimallman:
@kcranston These changes are on **devtree**. You might need to clear cache to see the auto-hiding behavior when changing nodes. I'm trying a basic blue for the nudge message.

PR #1092, opened 2016-11-05
New conflict display

jimallman:
This is client-side support for @jar398 's pull request https://github.com/OpenTreeOfLife/reference-taxonomy/pull/255:

 - Show our new-style node ids (e.g. 'mrcaott123ott456') if no taxon name is mapped to a node.
 - Remove pseudo-coloring of tips; instead, we show actual details from conflict service.
 - Replace '/synth@' in synth-tree links with '/@' to default to latest synthetic tree. 

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -6461,7 +6461,7 @@ function getNodeConflictDescription(tree, node) {
       case 'mapped_to_taxon':
           if (witnessURL) {
               conflictHTML = 'Aligned with <a href="'+ witnessURL +'" target="_blank">'+
-                  (node.conflictDetails.witness_name || "???") +'</a>';
+                  (node.conflictDetails.witness_name || node.conflictDetails.witness) +'</a>';
           } else {
               conflictHTML = 'Aligned with '+ missingWitnessDescription;
           }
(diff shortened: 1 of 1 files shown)


curator/static/js/study-editor.js
```diff
@@ -6436,11 +6421,21 @@ function getNodeConflictDescription(tree, node) {
...
+                    witnessURL = "/opentree/argus/{SYNTH_VERSION}@{NODE_ID}"
+                        .replace('{SYNTH_VERSION}', referenceTreeVersion)
+                        .replace('{NODE_ID}', node.conflictDetails.witness);
+
+                    * But this is not yet provided by the conflict service. Perhaps we could capture
+                    * this as <tree>.conflictDetails['referenceTreeVersion']
+                    *
+                    * For now, omitting the version entirely (e.g. '/@ott123') will redirect to
+                    * the latest-version URL.
+                    */
```
jimallman: @jar398 I could nail down these URLs if the conflict service would provide a version for the reference tree (e.g. 'opentree7.1' for the synthetic tree, or 'ott2.9' for taxonomy). Maybe it's not worth it; we'd probably need to wrap the existing conflict response in another layer to do this cleanly. Thoughts?

jar398: Not too hard, but I don't want to do this in a hurry and want to merge this PR this weekend. I'll make a separate issue for versions.

PR #1098, opened 2016-11-16
Return non-500 response if invalid NeXML uploaded.

### LINKED ISSUE #1097: Uploading an invalid NeXML file gives mysterious "reload" message.
I noticed this in other testing. In the Trees tab of the curation app, guessing wrong on the file format option can lead to a cryptic 500 error from the `to_nexson` API method, which in turn prompts web2py to display this message:

> An error occurred, please [reload] the page

This is a Bad Thing since it encourages the user to abandon work in progress over what is actually a minor and recoverable error.

jimallman:
This is to avoid a cryptic message that pushes the curator to reload the page (potentially losing work) for no good reason. The new exception (a 501 response) should give a friendly hint to examine the uploaded file more closely and choose the proper file format:

screen shot 2016-11-15 at 8 23 47 pm

Fixes #1097.

### DIFF
--- a/curator/controllers/default.py
+++ b/curator/controllers/default.py
@@ -678,7 +678,7 @@ def to_nexson():
                         dfj = get_ot_study_info_from_nexml(NEXML_FILEPATH,
                                                            nexson_syntax_version=NEXSON_VERSION)
                     except:
-                        raise HTTP(501, T("Submitted data is not a valid NeXML file, or cannot be converted."))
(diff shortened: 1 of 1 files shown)


jimallman:
After some discussion, we've decided this should return a 400 error "Bad request 400", rather than implying a server-side problem, see https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_Error

jimallman:
@jar398 The revised error code is live on **devtree** (behavior is as expected, so this looks good).

Issue #499, opened 2014-11-10
Synonyms used in mapping should be visible to the user in some way


mtholder:
For example in commit,
https://github.com/OpenTreeOfLife/phylesystem-1/commit/dfd9503d8706c33d814d253f0f455ef9c798b466

I modified 'Eriocran' to 'Eriocrania subpurpurella' then when I clicked the map, it resolved to "Dyseriocrania subpurpurella" in OTT.

That mapping is the correct. "Eriocrania subpurpurella" is a junior synonym (suppl. mat. of the paper confirms this).

But it would be nice if the user could see that the match was actually a perfect match, and the discrepancy is due to synonymy. The TNRS did not just make a "close enough" match of the text strings. 

In this case it is not obvious because the valid name is only 3 letters off from the junior synonym. But (due to taxonomists tendencies to reuse Greek/Latin roots to generate very similar genus names, and the preservation of the specific epithet in most reclassifications) this sort of case won't be very rare.

*edited to clarify that the "suppl. mat." refers to info from the supplementary materials of the publication.

mtholder:
@jimallman looks great to me. 
My (slight) preference would be that we would find a way to add the matched name (the junior synonym) to the nexson as an optional field.  That would make it easier for people who are working with the corpus of NexSONs to understand what happened. But that is a minor concern. Perhaps we can discuss it on the next call, and make an issue for that feature if we decide that it is wise.

jimallman:
Agreed, let's discuss and (if desired) make a separate issue for adding this to nexson. Keeping this open as a reminder to have that discussion. :grin:  Maybe discuss on the **opentree** (or **opentree-software**) group?

PR #1112, opened 2016-12-08
Show only one mapping candidate per taxon.

### LINKED ISSUE #860: Duplicate suggestions in OTU mapping (pick one?)
See for example this study (or force the same name in any study):
https://devtree.opentreeoflife.org/curator/study/edit/pg_2854/?tab=otu-mapping&match=melicy

Clear all mappings, set context to **All life** and fuzzy mapping = yes. Now look at the suggestions for taxon _Melicytus novae zealandiae_. Three suggested mappings, all pointing (for different reasons) to the same ottid. We should pare these down and present only the most sensible one.

screen shot 2016-01-22 at 12 05 35 pm

jimallman:
This simplifies the results in OTU mapping when the TNRS returns multiple "hits" on a single taxon (typically due to synonyms and alternate spellings). Fixes #860 and https://github.com/OpenTreeOfLife/taxomachine/issues/144; might also be relevant to #499.

These changes are working now on **devtree**. To test, try OTU mapping in any study with the following mapping options:
 - Search context is "All life"
 - Fuzzy search is active

Now customize any OTU's label to be "Homon" and map it to see multiple hits on different taxa. This behavior is unchanged.

Customize any OTU's label to "Melicytus novae zealandiae" and map it to see just one mapping candidate as shown here:
synonym-tooltip

We see the taxon's unique name, but the mouse-over text explains the synonym (alternate spelling) that was actually closest to our proposed label. 

To evaluate [our simple decision rules for choosing which mapping candidate](https://github.com/OpenTreeOfLife/opentree/commit/9625431460c32e7119d794d0e651c0cee9e48104#diff-ec50a802bd4aa427fc715b6e1989c6a2R6489) to show for a taxon, you can try mapping different names and compare the results on **devtree** vs. **tree** ("production"). You can also examine the raw TNRS response in your browser's Network tab or equivalent dev tool.

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -6481,18 +6481,70 @@ function requestTaxonMapping( otuToMap ) {
                     });
                     */
 
-                    var candidateMappingList = [ ];
-                    $.each(candidateMatches, function(i, match) {
-                        // convert to expected structure for proposed mappings
-                        candidateMappingList.push({
-                            name: match.taxon['unique_name'] || match.taxon['name'],       // matched name
-                            ottId: match.taxon['ott_id'],     // matched OTT id (as number!)
+                    /* TODO: If multiple matches point to a single taxon, show just the "best" match
+                     *   - Spelling counts! Show an exact match (e.g. synonym) vs. inexact spelling.
+                     *   - TODO: add more rules? or just comment the code below
+                     */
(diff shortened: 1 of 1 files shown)


jimallman:
> Now customize any OTU's label to be "Homon" and map it to see multiple hits on different taxa. This behavior is unchanged.

Apologies for the red herring. This example is just used to verify that we're not mixing candidates from different taxa.

PR #1121, opened 2017-03-03
Update displayed DOIs to match latest CrossRef guidelines

### LINKED ISSUE #1118: Use https: for DOI hyperlinks
See http://blog.crossref.org/2017/01/linking-dois-using-https-the-background-to-crossrefs-new-guidelines.html

Now remember that what someone sees on the web page, the hyperlink, and what we store in phylesystem are three different things. Not that I would object, but changing phylesystem seems painful, unnecessary, and possibly not compatible with software that reads the nexson. But the strings from the nexson can be converted from http: to https: before the HTML (or DOM) gets created.

Also, it looks like crossref, somewhat annoyingly, wants us to get rid of the 'dx' in the domain name: http://blog.crossref.org/2016/09/new-crossref-doi-display-guidelines.html
The 'new best practice' is e.g. https://doi.org/10.1629/22161
The two changes to the hyperlink URL can be made at the same time.

jimallman:
Fixes #1118 (see related discussion there). For backward compatibility, we preserve the old CrossRef format (**http://dx.doi.org/...**) in storage, but update to the latest (**https://doi.org/...**) in all displayed hyperlinks.

Note that there are three defined functions (two JS, one Python) that have the same behavior; if the CrossRef guidelines change again, we'll need to update them all!


jimallman:
This is working (and tested) on **devtree**, in both the main webapp (synth-tree viewer, About pages) and the curation app.

PR #1125, opened 2017-03-22
stats for OTT 3.0

jar398:


kcranston:
This is a duplicate, because I already updated both ott.json and
synthesis.json in a separate PR.

jar398:
Fixed node count (you have to subtract 1 for the header row)

PR #1128, opened 2017-05-09
use parens in taxo browser to distinguish merged from aligned qids

jar398:
This just tweaks the formatting of the source id list. Instead of

>ncbi:8397, worms:448307, gbif:6746, irmng:103657

it now shows

>ncbi:8397 (worms:448307, gbif:6746, irmng:103657)

which I hope is suggestive of the fact that the taxon comes from the non-parenthesized source. The others are just alignments.

See https://devtree.opentreeoflife.org/taxonomy/browse?id=364560 . Attempts to address https://github.com/OpenTreeOfLife/reference-taxonomy/issues/328

### DIFF
--- a/taxonomy/cgi-bin/browse.py
+++ b/taxonomy/cgi-bin/browse.py
@@ -296,7 +296,11 @@ def display_basic_info(info, output):
     # Sources
     start_el(output, 'span', 'sources')
     if u'tax_sources' in info:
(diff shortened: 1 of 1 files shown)


jar398:
Some discussion https://gitter.im/OpenTreeOfLife/public?at=5914bfba2b926f8a674694c5

kcranston:
I think this is an improvement over the existing display. As @jimallman noted, we could add some mouseover text but I am going to merge as is.

PR #1130, opened 2017-05-24
Fix entities in SVG download

jimallman:
We've had some problems with unsupported HTML entities (typically `&nbsp;`) in our SVG output. See this [related Gitter chat](https://gitter.im/OpenTreeOfLife/public?at=5925a2059f4f4ab05bf8002f) for details. 

This change serializes the main SVG node to Unicode, which removes all entities entirely, then relies on Base64 to encode all characters in the data URI. This is working and tested on **devtree**, see for example [this tree](https://devtree.opentreeoflife.org/curator/study/edit/tt_103?tab=trees&tree=tree3) with unlabeled internal nodes. This should download and print with no problems. Also note that I've added some diacritical marks and challenging punctuation in the topmost tip's label `Mycôsphærëlla "zoomer" >!;`. The new solution seems to handle these characters just fine.

### DIFF
--- a/curator/static/js/d3.phylogram.js
+++ b/curator/static/js/d3.phylogram.js
@@ -496,9 +496,11 @@ if (!d3) { throw "d3 wasn't included!"};
                       return '<a xlink:href="'+ link +'" target="_blank" xlink:title="'+ title +'">'+ d.name +'</a>';
                   case ('empty'):
                       /* N.B. empty label should hide, but still have layout to
-                       * support a legible highlight.
+                       * support a legible highlight. NOTE that we use the
+                       * numeric entity '#160' here, instead of 'nbsp', so we
+                       * can safely download the resulting tree view as SVG/XML!
                        */
-                      return '&nbsp; &nbsp; &nbsp; &nbsp;';
+                      return '&#160; &#160; &#160; &#160;';
                   default:
                       console.error('Unknown label type! ['+ d.labelType +']');;
                       return "???";
--- a/curator/static/js/d3.phylogram.js
+++ b/curator/static/js/d3.phylogram.js
@@ -500,7 +500,7 @@ if (!d3) { throw "d3 wasn't included!"};
                        * numeric entity '#160' here, instead of 'nbsp', so we
                        * can safely download the resulting tree view as SVG/XML!
                        */
-                      return '&#160; &#160; &#160; &#160;';
+                      return '&#160; &#160; TEST &#160; &#160;';
                   default:
                       console.error('Unknown label type! ['+ d.labelType +']');;
                       return "???";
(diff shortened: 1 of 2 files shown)


jimallman:
Should we need to explore heavier client-side libraries for decoding/encoding HTML entities to XML-compatible numeric entities, one promising option is https://www.npmjs.com/package/html-entities

josephwb:
Doesn't seem to work for me. I tried [this tree](https://devtree.opentreeoflife.org/curator/study/view/pg_11/?tab=trees&tree=tree9&conflict=ott) which gives the same error as before. Clicking on the utton in Safari brings up this new page:
tree svg

jimallman:
@josephwb These are working for me on **devtree**. Please try again after clearing your browser's cache. (This can be difficult in Safari. What version are you running?)

josephwb:
Ah, okay, I too can not get it to work on Safari. Works fine in Opera if that helps.

jimallman:
Thanks, it looks like XMLSerializer works differently in Safari, preserving HTML entities like `&nbsp;` and `&gt;` (in tip label `Mycôsphærëlla "zoomer" >!;` above). I'll look at some more thorough fixes and update this thread when there's something to try.

jar398:
Sounds like a bug in Safari XMLSerializer . The only mention of it I found was https://github.com/pbakaus/domvas/issues/11 which is rather muted. I wonder whether there is a good way to report it.

jimallman:
> Sounds like a bug in Safari XMLSerializer

Not so much a bug as a subtle but intended behavior, apparently. The doctype of the chosen node's parent document (in this case, the surrounding web page) is used to determine how entities are treated, as [described in this WebKit test](https://github.com/WebKit/webkit/blob/5277f6fb92b0c03958265d24a7692142f7bdeaf8/LayoutTests/fast/dom/XMLSerializer-serializeToString-entities-expected.txt). In short, if the parent doctype is XHTML, serialization will use HTML entities that will fail in "vanilla" XML documents.

Since `document.doctype` is a read-only property, the solution is to clone the SVG node and wrap it in a new, temporary XML document. Since [DTDs are no longer recommended for SVG docs(!)](https://jwatt.org/svg/authoring/#doctype-declaration), we rely on `document.implentation.createDocument()` to create a "vanilla" XML doc if no doctype is provided.

This works for me on **devtree**, including in Safari, for both [my preferred tree](https://devtree.opentreeoflife.org/curator/study/edit/tt_103?tab=trees&tree=tree3) and [the one @josephwb is attempting to use](https://devtree.opentreeoflife.org/curator/study/view/pg_11/?tab=trees&tree=tree9&conflict=ott). I'll keep testing in other browsers, but please let me know if you see more failures on **devtree**!

Issue #726, opened 2015-08-28
Explain/mention that some non-binomials can be mapped to OTT


rgazis:
In some cases, terminals comprising on genus + strain # can be mapped to OTT. 
![screen shot 2015-08-28 at 2 11 33 pm](https://cloud.githubusercontent.com/assets/6391539/9553726/fcc25ece-4d8e-11e5-9e82-b5aaa6e0f419.png)

Can be confusing to authors, and it might be good to mention in the tutorial/manual in which cases non-binomials can be mapped.

jimallman:
Am I correct that this refers to names like "Lecanicillium sp. CBS 100890"? What would be the desired mapping in this case?

jar398:
I think the question is less about what the mapping is (in this case it's quite clear), than about the user wanting some model of when they'll find something and when they won't.

About the only thing to say about this is that a non-binomial like this is in OTT only if NCBI Taxonomy has it, and whether NCBI Taxonomy has it has to do with the vagaries of curation of the Genbank record. So it is quite spotty and out of our control. The policy for strains in the taxonomy has recently changed; NCBI is phasing out strains as "taxa" in their taxonomy.

jar398:
This is probably oversharing, but for most molecular studies, all of the sequences are in Genbank, which means that all the tips are mappable. Assuming the sequences were deposited prior to the NCBI Taxonomy capture used in the latest OTT, if the TNRS doesn't give a mapping, you ought to be able to find a table among the supporting material for the publication that lists all of the samples, with Genbank id and something resembling the tip label. You can then look up the Genbank id in nuccore, and the nuccore record will always link to an NCBI taxon record. You can then feed the name in that record to the TNRS to get a good OTT mapping for the tip. I bet $1 this would work for the failed mapping in Romina's example.

I've done this and it works, but it's incredibly labor intensive.  You sort of wish they had just put the Genbank id in the tip label. If they did that, we could write a script to look them all up. But of course this would not be natural given the authors' workflow - they do the phylogenetic analysis and prepare the Newick before they do the Genbank deposits.

Alternatively, the curator could upload the tip table from the supporting material, and we could parse it to find the Genbank ids, and so on. But usually the table is in a PDF, not in any useable form, so this usually this will just be a dream.

jimallman:
Food for thought, thanks @jar398!

jimallman:
[...]

PR #1131, opened 2017-06-01
Hint about mapping strains

jimallman:
Explain that tips with genus+strain might be mappable. This adds text to the initial display, and a persistent hint in the Help box for OTU Mapping. Addresses #726.


jimallman:
See discussion in #726 for screen grabs of new text.

Issue #1024, opened 2016-08-15
curator app: if some branches have lengths and others don't, the app lies to the user.


mtholder:
see https://tree.opentreeoflife.org/curator/study/view/ot_760

jimallman:
My bad -- the code currently just checks a single branch (first one found), assuming that branch lengths would apply to all-or-none. Am I correct that the misrepresented tree above is the one that currently shows as "no branch lengths found"?

@mtholder (or others), please describe the proper behavior. Should I check _all_ branches for length, and treat a single instance as a tree-with-branch-lengths? If so, how should the different branches (with and without length) be represented visually in the tree view? Should they be distinct in color or style, or is it enough that the length figure (red) only appears on branches with length?

mtholder:
I think that we need to check all. if any are missing lengths, then I think that we can disable the tree-with-branch-length view. I think we just want to say "Some of the branches are missing lengths" rather than "no branch lengths found".
I don't *think* is not a common case, but it is disconcerting to users who know that some branches have lengths.

jimallman:
Thanks, will do!

jimallman:
@mtholder In cases where some-but-not-all branches have length, I assume we still want to gather 'Branch length mode', and possibly time units. Correct?

mtholder:
Yes. I think we'd want the user to be able to specify a units for the branches that have lengths.

jimallman:
See the linked PR above for details on the proposed fix, and a test URL on **devtree**.

PR #1134, opened 2017-06-08
Careful branch length detection

jimallman:
As discussed in #1024, I've been playing too fast and loose with branch-length detection. We need to more carefully distinguish when a tree has all/some/no edges with lengths assigned, since each case has different consequences for UI and behavior in the single-tree popup.

 * If **all edges** have lengths:
    * prompt for related properties
    * enable the tree-view toggle to hide/show branch lengths

 * If **some (but not all) edges** have lengths:
    * prompt for related properties 
    * disable the tree-view toggle

 * If **no edges** in the tree have lengths:
    * don't prompt for related properties 
    * disable the tree-view toggle

Since the different cases and their effects aren't immediately obvious, I've added stronger hints to the branch-length toggle when it's disabled, for example:

toggle-hints

These changes are live on **devtree**. You can see the effects of each case using [the specially prepared trees in this study](https://devtree.opentreeoflife.org/curator/study/edit/tt_103?tab=trees).

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -1632,27 +1632,57 @@ function toggleRadialTreeLayoutInViewer(cb) {
     /* N.B. This logic is replicated in Knockout bindings to handle the 
      * initial display for each tree!
      */
-    if (usingRadialTreeLayout || !allBranchLengthsFoundInTree(currentTree)) {
+    if isBranchLengthToggleEnabled(currentTree) {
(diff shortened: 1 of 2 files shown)


jar398:
I think this was my fault - I remember advising you not to worry about this problem. I wonder how it comes about (upstream) that some but not all edges have lengths? But that's not an important question, I guess, since we now know that it happens.

PR #1136, opened 2017-06-20
Cache TNRS repeats

### LINKED ISSUE #1046: curator app: OTU mapping enhancement. Cache the TNRS results for a "modified for mapping" string
ot_800 is huge and has a lot of barcoding data leading to a lot of sub-specific sampling.

After I apply a regex to the names, it is often the case that the "modified for mapping names" become identical to some other "modified for mapping name".

In a really big study like this one, it would be nice if the TNRS call for those names was just done once. I didn't look at the code, but from the UI it looks like the same name can get sent multiple times.

I doubt this is terribly common, so I would not call this a very high priority.

jimallman:
Refactor TNRS query code to use cached responses (stashed in a local JS object) where the TNRS request payload is identical to one previously cached. This handles common cases where many "modified for mapping" labels end up asking for the same (e.g.) species. Fixes #1046. 

Note that the local TNRS cache will be cleared if the curator successfully adds new taxa in the OTU Mapping UI, since this can immediately result in different TNRS responses for the remaining OTUs.

This is working (and fairly thoroughly tested) on **devtree**.

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -6466,14 +6466,14 @@ function requestTaxonMapping( otuToMap ) {
         url: doTNRSForMappingOTUs_url,  // NOTE that actual server-side method name might be quite different!
         type: 'GET',    // to allow caching for identical requests!
         dataType: 'json',
-        data: {
+        data: JSON.stringify({
             "names": [searchText],
             "include_suppressed": false,
             "do_approximate_matching": (singleTaxonMapping || usingFuzzyMatching) ? true : false,
             "context_name": searchContextName
-        },  // data (asterisk required for completion suggestions)
+        }),  // data (asterisk required for completion suggestions)
         crossDomain: true,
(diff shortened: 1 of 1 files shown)


jimallman:
Any thoughts on whether this cache needs to be size-limited, perhaps using a FIFO approach? The returned data is pretty tiny, so I'm thinking the cache will be tiny compared to even a typical study, but maybe that's optimistic if someone is working with thousands of OTU labels...

jimallman:
> Any thoughts on whether this cache needs to be size-limited, perhaps using a FIFO approach?

I've gone ahead with a dirt-simple FIFO system that will toss the first key when the cache size (number of items) exceeds a preset `TNRSCacheSize`, currently set to 200. I believe this will handle the expected case of many labels being modified to the same string.

jimallman:
@mtholder If this looks like a good solution, please review and merge.

PR #1140, opened 2017-12-21
Update About > Bibliographic References fetch

jimallman:
This started failing (not clear why) on the latest **devtree** (after some system updates). It seems we now need to be more careful about using gluon's fetch tool, or (as I've done here) replace it with our preferred `requests`.


jimallman:
This is [working now](https://devtree.opentreeoflife.org/about/references) on **devtree**.

PR #1141, opened 2017-12-21
Fixes to broken subtree-download feature

jimallman:
This is now breaking on devtree. I'm betting a move from gluon.tools.fetch to requests will fix it.

(Yes! This is now working on **devtree**, see for example [this node](https://devtree.opentreeoflife.org/opentree/argus/opentree9.1@ott678209/Nanoarchaeota). REMINDER: Click **Download subtree as Newick string** to download the subtree as Newick.)

### DIFF
--- a/webapp/controllers/default.py
+++ b/webapp/controllers/default.py
@@ -99,8 +99,12 @@ def download_subtree():
      
     try:
         # fetch the Newick tree as JSON from remote site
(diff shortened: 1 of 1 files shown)


jimallman:
There's also an incidental change to use json (instead of simplejson) in another method here.

PR #1143, opened 2018-04-13
Show server-side TNRS error in the synth-tree view

jimallman:
This makes it clear in the (hopefully rare) case that TNRS is choking on
a searched name, and any information we can provide about why. Addresses https://github.com/OpenTreeOfLife/feedback/issues/383.


jimallman:
See OpenTreeOfLife/feedback#383 for example screenshots of the new (visible) error messages.

mtholder:
looks good to me. thanks @jimallman

PR #1146, opened 2018-04-17
Fix broken curator profile page (failed to load)

### LINKED ISSUE #1145: Curator profile page is broken (on devtree)
I will chase this Real Soon Now...

jimallman:
Fixes #1145. This also has a nice compact display of studies added/curated/in-synthesis, but I'm commenting out the new markup until we're using otindex (since oti still reports just one curator name per study).

### DIFF
--- a/curator/controllers/study.py
+++ b/curator/controllers/study.py
@@ -100,7 +100,11 @@ def _get_latest_synthesis_details_for_study_id( study_id ):
             # Prepend scheme to a scheme-relative URL
             fetch_url = "https:%s" % fetch_url
         # as usual, this needs to be a POST (pass empty fetch_args)
-        source_list_response = requests.post(fetch_url, headers={"Content-Type": "application/json"},data=simplejson.dumps({'include_source_list':True})).text
+        source_list_response = requests.post(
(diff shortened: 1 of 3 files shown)


curator/controllers/study.py
```diff
@@ -100,7 +100,11 @@ def _get_latest_synthesis_details_for_study_id( study_id ):
             # Prepend scheme to a scheme-relative URL
             fetch_url = "https:%s" % fetch_url
         # as usual, this needs to be a POST (pass empty fetch_args)
-        source_list_response = requests.post(fetch_url, headers={"Content-Type": "application/json"},data=simplejson.dumps({'include_source_list':True})).text
+        source_list_response = requests.post(
+            fetch_url,
+            headers={"Content-Type": "application/json"},
+            data=simplejson.dumps({'include_source_list':True})
```
bredelings: Is it possible to switch from simplejson to json here?

jimallman: Certainly. I'll make that change in all three instances and update the PR. Thanks for looking out!

jimallman:
@bredelings Any other concerns here?

PR #1147, opened 2018-04-19
Remove last few vestiges of simplejson in webapps

jimallman:
We stick to the built-in `json` module now, which has the same API. All changes tested on local dev server. Addresses comment in #1146.


jimallman:
This is deployed (and apparently working fine) on **devtree**.

PR #1148, opened 2018-05-07
Add OT Taxonomy citation to About page

jimallman:
This follows the convention set by the main Open Tree of Life citation.
Addresses #1137.

This is [available for review on **devtree**](https://devtree.opentreeoflife.org/about/open-tree-of-life).

NOTE that I've added quotation marks around the title, following the example of the main OToL paper citation on this page. Is this OK?

Also, note that I've removed the `+` from the main OToL citation above, since this looks to me like a typo. Is that correct?


kcranston:
Looks good to me. Thanks, @jimallman

Issue #1138, opened 2017-07-16
Bottom of treeview gets cut off


snacktavish:
e.g. https://tree.opentreeoflife.org/opentree/opentree9.1@ott471706/Rotifera
the bottom half of words at the bottom of the view is cutoff in both firefox and chrome, independent of window size or scrolling.

jimallman:
Ah, I see the problem. This happens when a "more..." cluster (for hiding excessive child nodes) has only one child node inside. The cluster itself is taller than a single node, so we don't allow adequate height in the argus view. I'll modify the logic so that we don't bother creating single-child clusters, which are dumb for a few reasons.

jimallman:
Fix is on **devtree**, moving these changes to a new pull request...

PR #1150, opened 2018-05-07
Argus view fixes

jimallman:
These fixes address #1138, and a related bug I found when clusters have a very wide label (text was being "cropped" on mouseover).

### DIFF
--- a/webapp/static/js/argus/drawtree.js
+++ b/webapp/static/js/argus/drawtree.js
@@ -165,8 +165,8 @@ function createArgus(spec) {
         "currDisplayContext": undefined, //arg to most recent displayNode call
 
         clusters: {},      // a registry of clustered child-nodes, keyed to the parent-node's ID
-        clusterSize: 100,   // try to bundle sets of n children
-        minClusterSize: 10, // add fewer to the previous cluster instead
+        maxClusterSize: 100,   // try to bundle sets of n children
+        minClusterSize: 10, // add fewer to the previous cluster (or unclustered children) instead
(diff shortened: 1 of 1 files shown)


jimallman:
@snacktavish Please take a look on **devtree** and (if satisfied) merge this PR and close #1138 (see comments there for a concise explanation).

jimallman:
In short, this bug was a result of too-small clusters (fewer than our designated 10 nodes) being created in the argus view.

snacktavish:
Looks good - except I just realized that names are getting similarly cut off on the left as well, even though there is plenty of room (on the right). Should I merge this PR and make a new issue for that, or is it more straightforward to combine here?

jimallman:
OK, it looks like we need to build the entire tree display, _then_ reckon its bounding box and resize the canvas to match. I'll need to test the zoom/magnify code to see if we can still maintain the center of the viewport.

> Should I merge this PR and make a new issue for that, or is it more straightforward to combine here?

I'll try to add this to the current PR, so please stand by...

jimallman:
OK, I believe this is working as expected on **devtree**. We now reckon the actual bounding box of all visible tree elements, resize the SVG canvas to include them all, and update the scrollbars to maintain our previous position (based on upper-left corner) as closely as possible. Note that we also now echo the full node label on mouse-over, for cases when a dense tree is partially out of view:
node-label-title-grab

This adds a little visual clutter, but it might help someone to quickly scan a bunch of similarly-labeled siblings. Thoughts?

Anyway, please take another look and merge if it looks good!

snacktavish:
beauuutiful!

PR #1152, opened 2018-05-23
Nexttree

mtholder:
Main new feature: Use a config-based setting to generate links to the curation application. Useful if we want to just launch a tree browser. Still need to new work configuration into deployment. If the config variable is copied over to from the example, the default is old behavior (assume that curator is running on the same machine as the webapp).


jimallman:
NB - This is what's currently running now on **ot37**.

mtholder:
docs probably need a relook at this point. closing and leaving branch

PR #1153, opened 2018-05-23
Allow more recent stevedore.

mtholder:
Not sure how to test if this change works. I see no
obvious problems with it. I don't even see where we
use this package within our webapps, so we can probably
get rid of it.

The version 0.8 that we were previously pinned to does
not install cleanly on any modern Python version that I've seen.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 nose==1.2.1
 ipython==0.13.1
 wsgiref==0.1.2
-stevedore==0.8
+stevedore>=0.8
 pyparsing==1.5.7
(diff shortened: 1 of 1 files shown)


jimallman:
Indeed, it looks like this was [included (by me!) in our initial requirements.txt](https://github.com/OpenTreeOfLife/opentree/commit/78962f9f94774723dad1aab328a0a7d9a3df05d5) file a few years ago. This initial set was probably just the output of `pip freeze`, so it's safe to say that stevedore was picked up accidentally. I suspect we can simply get rid of it.

Issue #1158, opened 2018-06-07
Subtree download (Newick) fails for some nodes in synthetic tree


jimallman:
Reported by @blackrim [on gitter](https://gitter.im/OpenTreeOfLife/public?at=5b19477c22a05f5e710b3668).

jimallman:
Testing with chatty production logs identifies the problem as a "poison pill" in some Newick results, probably Unicode (non-ascii) characters in some node labels. It seems we're [coercing the `newick` member of a subtree response to `str` here](https://github.com/OpenTreeOfLife/opentree/blob/e8f7069f4ca154933ff68d58d03d9b242c29cc5a/webapp/controllers/default.py#L105), when instead we should use `unicode`. This neatly resolves the issue in the nodes I've tested. I'll move this change from production back to source and build a pull request.

PR #1159, opened 2018-06-07
Subtree download with unicode chars

jimallman:
This addresses the reported problems with subtree downloads in #1158. The simple fix is tested and (apparently) working now on **devtree**, for example [in the reported Ericales](https://devtree.opentreeoflife.org/opentree/argus/opentree10.2@ott648892/Ericales).


mtholder:
thanks for tracking that down, @jimallman

PR #1163, opened 2018-07-06
sanitize XML with bleach

mtholder:
I think this is all that is needed to avoid the XSS stuff coming through via comments, but I suppose we should check it more thoroughly... github users named <script>... etc

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 requests==2.2.1
 python-dateutil==2.2
 locket==0.1.1
+bleach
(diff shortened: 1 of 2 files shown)


snacktavish:
Looks good to me, or at least like a good start!

mtholder:
OK. I'll go ahead and merge it and deploy. A couple of notes (in case we revisit). Using `sanitize=True` or `unicode` instead of `str` in the changed line seem more appealing. Both result in exceptions when used inside web2py's DIV function. So, bleach.clean seems the best easy option.

mtholder:
I'm afraid that I don't have a log of the exception from the `sanitize=True`...
I'll try to look over the https://github.com/OpenTreeOfLife/opentree/pull/1180 I assume that is what you referring to by the belt+suspenders.

jimallman:
OK, thanks! So far I'm not seeing any exceptions from sanitize=True, after testing with a variety of comments and version notes. So I'm include to leave it in place and respond if something goes :boom: :boom: .

PR #1170, opened 2018-08-28
usage of new NCL option

mtholder:
this should limit memory usage for conversion of NEXUS files with huge character matrices. We discard the character data in the NeXML -> NEXSON stage, but are currently including it in the NEXUS->NeXML stage. This commit fixes that. I'm going to merge, deploy and test now....


mtholder:
test on the deployment, that is. I have tested the new NCL option...

PR #1179, opened 2018-12-22
Improved html bleaching

jimallman:


jimallman:
Clobbering this, going to try again after updating development to get recent commits on master.

PR #1180, opened 2018-12-22
Improved HTML bleaching

jimallman:
Use custom tag and attribute whitelists to allow selected markup in feedback/comments and in version notes for synthesis and taxonomy. We also re-enable the `sanitize=True` option for web2py's `XML` helper, for extra protection; this is overkill, but it doesn't cause any visible harm.

### DIFF
--- a/webapp/controllers/about.py
+++ b/webapp/controllers/about.py
@@ -4,6 +4,24 @@ from datetime import datetime
 from opentreewebapputil import (get_opentree_services_method_urls, 
                                 fetch_current_TNRS_context_names,
                                 get_data_deposit_message,)
+import bleach
+from bleach.sanitizer import Cleaner
+
+# Define a consistent cleaner to sanitize user input. We need a few
+# elements that are common in our markdown but missing from the Bleach
+# whitelist.
+# N.B. HTML comments are stripped by default. Non-allowed tags will appear
+# "naked" in output, so we can identify any bad actors.
+common_version_notes_tags = [u'p', u'br',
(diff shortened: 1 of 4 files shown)


jimallman:
These changes are live on **devtree**. For example, see Bacteria in the synth-tree viewer for a comment with lots of markup and a naughty SCRIPT element. Version notes in the synth-tree viewer (About > Latest ... version) should also show nicely formatted tables.

jimallman:
Thanks for catching this, @snacktavish! It looks like web2py's builtin sanitizer fails when handling Unicode. See [this web2py issue](https://github.com/web2py/web2py/issues/1624) and the subsequently merged [pull request](https://github.com/web2py/web2py/pull/1625), which was only applied in a more recent version (2.17.2) of web2py than the one that we use (2.8.2). Rather than introduce this as a patch, I'll try relaxing our current "belt and suspenders" approach and rely on just the bleach library.

jimallman:
OK @snacktavish, your links above are now working as expected on **devtree**. Sadly, your multi-line 
ASCII cow is goofy, since I haven't whitelisted the `PRE` element. Let me know if you'd like to see this or other elements approved!

jimallman:
@snacktavish Yes,  OpenTreeOfLife/phylesystem-api#218 should also be merged. It uses the bleach library to sanitize curation comments, so it's just "suspenders" I guess. (We no longer trust the original "belt", web2py's sanitize function, since it doesn't handle Unicode.)

snacktavish:
done!

Issue #1186, opened 2019-04-07
url encode a request to show the legend


mtholder:
while writing some docs for opentreeoflife.org it occurred to me that it would be nice to have a URL either for the legend for navigating the tree (the drop overlay that explains dashed lines for taxonomy) or `show_legend=true` URL parameter to cause the overlay to be shown on any page.

jimallman:
Closing this (related PR already merged).

PR #1188, opened 2019-04-12
Support URL option to show synth-tree legend automatically.

jimallman:
If someone adds `?show-legend=true` to any synth-tree viewer URL, the legend popup will appear automatically when the page loads. Addresses #1186.

Screen Shot 2019-04-12 at 1 36 19 PM

Note that synth-tree URLs typically go through a redirect to match the latest synthesis, so this addition to the query-string will be recognized, but won't be in the final URL displayed in the browser.


jimallman:
This is working now on **devtree**, see for example http://devtree.opentreeoflife.org/opentree/opentree10.4@ott332573/Opisthokonta?show-legend=true

PR #1190, opened 2019-04-17
Improved banner for development sites.

jimallman:
The "DEVELOPMENT" banner will now appear in **any** domain other than **tree.opentreeoflife.org**, instead of just **devtree**. This should cut down on confusion when users stumble into other dev systems like **ot123.opentreeoflife.org**.

Another enhancement: Instead of simple rollover text, we expand the entire banner on mouseover. This gives more room for explanation, plus a live (clickable) link to the production website:

Screen Shot 2019-04-16 at 8 00 13 PM

This addresses a recent issue or discussion... that I can't seem to find now. (Link welcome, if anyone spots it!).

### DIFF
--- a/curator/static/css/default.css
+++ b/curator/static/css/default.css
@@ -773,9 +773,17 @@ body.printing-tree-view #tree-title * {
     color: #888;
 }
 /* Warning banners appear when using DEVELOPMENT, STAGING servers */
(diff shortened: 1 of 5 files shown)


jimallman:
This behavior is in place now on **devtree**. Note that this requires the latest CSS, so you might need to clear your browser's cache if it looks funny.

Issue #1195, opened 2019-07-09
change irmng link outs


mtholder:
Not sure if this is opentree repo or another...but see:

https://gitter.im/OpenTreeOfLife/public?at=5d241bf7b0027d2b199b9b05

jimallman:
I'll chase this down (in the webapps, at least) and report back here.

jimallman:
@mtholder Can we close this issue? I've merged the related PR, so all should be well (except a few [outdated IRMNG ids](https://github.com/OpenTreeOfLife/reference-taxonomy/issues/362)) next time we deploy to production.

PR #1196, opened 2019-07-09
irmng url fix

mtholder:
fixes issue-1195

### DIFF
--- a/taxonomy/cgi-bin/browse.py
+++ b/taxonomy/cgi-bin/browse.py
@@ -322,7 +322,8 @@ def source_link(source_id):
             elif parts[0] == 'gbif':
                 url = 'http://www.gbif.org/species/%s/' % parts[1]
             elif parts[0] == 'irmng':
-                url = 'http://www.marine.csiro.au/mirrorsearch/ir_search.taxon_info?id=%s' % parts[1]
+                # url = 'http://www.marine.csiro.au/mirrorsearch/ir_search.taxon_info?id=%s' % parts[1]
+                url = 'http://www.irmng.org/aphia.php?p=taxdetails&id=%s' % parts[1]
             elif parts[0] == 'if':
                 url = 'http://www.indexfungorum.org/names/NamesRecord.asp?RecordID=%s' % parts[1]
             elif parts[0] == 'worms':
--- a/webapp/static/js/treeview.js
+++ b/webapp/static/js/treeview.js
@@ -988,7 +988,7 @@ function showObjectProperties( objInfo, options ) {
                                 break;
 
                             case 'IRMNG':
-                                displayVal = '<a href="http://www.marine.csiro.au/mirrorsearch/ir_search.taxon_info?id='+ sourceInfo.taxSourceId +'" '
+                                displayVal = '<a href="http://www.irmng.org/aphia.php?p=taxdetails&id='+ sourceInfo.taxSourceId +'" '
                                               + 'title="Interim Register of Marine and Nonmarine Genera" target="_blank">IRMNG: '+ sourceInfo.taxSourceId +'</a>';
                                 break;
 


jimallman:
I just tested these URL changes (manually), and it seems the IRMNG ids might also be changing..? Here are the old and new URLs for bacteria (IRMNG id=13):

OLD (works): http://www.marine.csiro.au/mirrorsearch/ir_search.taxon_info?id=13
NEW (fails): http://www.irmng.org/aphia.php?p=taxdetails&id=13

I think (apologies if I'm linking to the wrong "bacteria") its new IRMNG id is 6:
http://www.irmng.org/aphia.php?p=taxdetails&id=6

@TonyRees, @mtholder, @jar398, @hyanwong (found in linked Gitter conversation in #1195): Please take a look. I hope I'm wrong here, but I'm hesitant to merge this PR until we're sure we don't need to update all our IRMNG ids.

jimallman:
OK, another test is somewhat reassuring. Centrohelida resolves in both the old and new URL:

http://www.marine.csiro.au/mirrorsearch/ir_search.taxon_info?id=11213
http://www.irmng.org/aphia.php?p=taxdetails&id=11213

So maybe Bacteria is a special case? If so, how many others are out there?

TonyRees:
Hi Jim, all,

Well, that was a new one on me! It turns out that when the data were ported
to VLIZ after the last CSIRO-origin dump in 2014 (the one used by OTT),
they made an attempt to revise IRMNG IDs for kingdoms (only) to match those
already in use in Aphia (WoRMS), resulting in the following changes:

Name rank IRMNG ID 2014 IRMNG ID 2019 note
Animalia kingdom 11 1
Archaea kingdom 12 8
Bacteria kingdom 13 6
Fungi kingdom 14 4
Plantae kingdom 15 3
Protista kingdom 16 14
Unaccepted kingdom 17  - dropped
Unallocated kingdom 18  - dropped
Viruses kingdom 19 10


For other ranks i.e phylum and below, IDs are unchanged (I just checked the
first few hundred, all OK).

There are a few names (mainly obsolete higher taxa at the specified rank)
since deprecated. Their IDs are retained on the system and a "proforma"
taxon page is returned by the current implementation with a pointer to the
current equivalent name and ID, like this (example: IRMNG
142 Chemoautotrophic Bacteria),
http://www.irmng.org/aphia.php?p=taxdetails&id=142:
----------------
IRMNG_ID: 142 has been DELETED  - reason: obsolete term
Please replace it by IRMNG_ID: 6 (Bacteria
<http://www.irmng.org/aphia.php?p=taxdetails&id=6>)
----------------

So... maybe a manual fix-up is required of the 9 (actually 7 still in use)
IRMNG names/IDs at kingdom level, on the version that you currently have
only. Note that post-2014, some new kingdoms have been introduced (most
notably Chromista) but these will not concewrn you until you do a new data
import.

It also turns out that IRMNG kingdom IDs 12 and 14 have actually changed
between 2014 and the current - this should not have happened but too late
now... 12 was Archaea, now Biota incertae sedis; 14 was Fungi,
now Protista. Again, once you have a fresh data dump these issues will
disappear.

The present (2016 onwards) IRMNG data dumps do not include species, since
the latter are getting outdated and generally have not been revisited since
either 2012 or 2006 in many cases. However some of these have been updated
e.g.moved to new parents in a small number of cases, plus the odd error
corrected. However the names and IDs should still match what you have
(presuming they are just mapped to an OTT name/ID) and maybe it is OK to
keep the old ones and just update the genus names and above. Alternatively
you could request a more extensive/custom data dump from the present VLIZ
custodians including species as well; however then that would not match the
citable, archived versions (genera-only) which are provided via
http://www.irmng.org/export/ . Something for you to consider before your
next update...

I think this covers everything. I will copy this to the gitter issue as
well unless it goes there automatically somehow.

Regards - Tony






Tony Rees, New South Wales, Australia
https://about.me/TonyRees

jimallman:
I've created a separate issue for the IRMNG-kingdoms quirk, since that's unrelated to the main issue here. (Fixes to the new issue will also fix the handful of bad IRMNG ids that appear here.)

Merging this now, since it's already better than the current behavior.

Issue #1193, opened 2019-06-07
Odd doi switch on study creation


snacktavish:
Creating a study from the url https://link.springer.com/article/10.1007/s11557-015-1098-z auto-populates a quite different paper, on sneutrinos, whatever those are (https://iopscience.iop.org/article/10.1088/1475-7516/2012/10/015/meta). It also updates the doi to the doi of this incorrect paper. 
If you create the study, and then fix the title, and search for doi, you get the correct doi.  (problem noted by @thomasstjerne )

jimallman:
We discussed this, and it seems the webapp accepts *either* a DOI or URL here, so we're probably mistaking this URL for a proper DOI. Hilarity ensues.

I'll take a closer look at this test. When in doubt, we should assume the submitted string is a URL and not a DOI.

jar398:
DOIs all start with '10.' Or 'doi:10.'

jimallman:
Understood. Nonetheless, we have assumed that a URL including `10.` has a complete DOI as the remainder of the URL. 

In this case, that's correct --`10.1007/s11557-015-1098-z` is [the DOI for a paper in _Mycological Progress_](https://search.crossref.org/?q=10.1007%2Fs11557-015-1098-z). In tracing the code, I realized that in this instance we're using [an outdated version of the search API](https://github.com/OpenTreeOfLife/phylesystem-api/blob/fd373ba3c9497f1e04d446d8478baceb3c803764/controllers/default.py#L1671) on crossref.org. Their current API [gives the expected result](https://api.crossref.org/works/10.1007%2Fs11557-015-1098-z), albeit in a more complicated wrapper. The old API works for most papers, so I'm guessing they stopped indexing new stuff to it at some point.  I'll just need to update this call to use the [new API and response format](https://github.com/CrossRef/rest-api-doc), and double-check [any other places where we call crossref.org](https://github.com/search?q=org%3AOpenTreeOfLife+%22search.crossref.org%22&type=Code) and update them as well.

jimallman:
A few other links of interest:

- Test the **latest results for a DOI** using the current [CrossRef web form](https://search.crossref.org/?q=10.1007%2Fs11557-015-1098-z)
- [CrossRef API status page](https://status.crossref.org/) (look here first, if API fails in the future)
- [API guidelines]( https://github.com/CrossRef/rest-api-doc#api-overview
) regarding whether to URL-encode slashes, etc. within a DOI _(short answer: yes)_

[...]

PR #1199, opened 2019-08-05
Overhaul of DOI lookup using latest CrossRef APIs

jimallman:
Reminder: This feature in the curation app uses existing reference text to look up a likely DOI for a publication. Addresses #1193.

This is tested and working on **devtree**, see for example [this Unicode-heavy reference](https://devtree.opentreeoflife.org/curator/study/edit/tt_241).


snacktavish:
Looks great! I tried a few different URLs and doi's and it behaved well.

PR #1204, opened 2019-08-19
Fix broken study list w/ multiple curators

jimallman:
We now cleanly split multiple curator names, and each becomes a separate
link to filter for that person's studies.


jimallman:
These changes are live on **devtree**. 

Personally, I'd prefer the curator-name links to go to each curator's profile page, instead of filtering the study list to show their contributions. But this would require some fancy dancing, since our study Nexson doesn't currently store a curator's Github userid (login), just their username (display name).

snacktavish:
Ya - getting it to go to the curator profile page at some point will be good, but this looks great for now!

PR #1206, opened 2019-09-06
Fix missing progress chart (and possibly others)

jimallman:
Failing to initialize this JS var is causing other client-side code to fail. This set a harmless default value on all pages. 

This fix is [working now on **devtree**](https://devtree.opentreeoflife.org/about/progress).

Thanks to @kcranston for the heads-up!


kcranston:
Awesome! Thanks.

PR #1209, opened 2019-10-13
hotfix to issue @hyanwong pointed out on gitter

mtholder:


mtholder:
tested and works on devtree

PR #1210, opened 2019-10-15
Fix "dubious data" warning: add "ott" prefix to OTT_VERSION

bredelings:


bredelings:
Checked that this fixes the problem on devtree.

PR #1219, opened 2020-02-01
Collection management UI - fixes and refresh feature

jimallman:
This fixes some serious bugs in the current collections editor, and adds the ability to update trees from phylesystem. 

Specifically, this will adopt any new tree labels and mark them for review. Same for trees (or studies) that have since been removed from phylesystem -- these are marked for likely removal from the collection, but also as hints that a replacement tree might be needed here.

Here's the resulting display after a "boring" update with no changes to tree labels or status:
Screen Shot 2020-01-31 at 7 55 15 PM

This one has more interesting changes to review:
Screen Shot 2020-01-31 at 7 42 28 PM

Note that we don't currently flag changes to _tree topology_, which might be very useful. We could approximate this by storing and updating a SHA for each tree, but this would mean that _any_ changes in the study would prompt for a review, so maybe not a good idea.

Anyway, this is deployed now on **devtree**. Please take a look and see how it behaves for you. [The example collection above](https://devtree.opentreeoflife.org/curator/collections/jimallman/new-test) is a good one to test, but please don't save your changes until others have had a little time to try it. The big input-to-synthesis collections are also interesting to look at. Generally their trees have not been removed, but tons of new+improved tree names can be incorporated easily now.

### DIFF
--- a/curator/static/css/default.css
+++ b/curator/static/css/default.css
@@ -932,3 +932,7 @@ a.node-direct-link {
     margin-right: -16px;
     border-radius: 4px;
 }
+.tree-status-marker {
+    font-weight: bold;
+    margin: 4px 0 0 2px;
+}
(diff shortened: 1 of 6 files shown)


jimallman:
Dangit, there's a [Known Issue in otindex](https://github.com/OpenTreeOfLife/otindex/issues/48) that might pop up in your testing. This might show a "new" empty tree name for named trees. Please disregard this behavior; it should be fixed Real Soon Now

PR #1221, opened 2020-02-17
Highlight extinct nodes in synth view

jimallman:
This reflects our recent discussions about how to display extinct nodes in the synthetic tree. Nodes and "rootward" paths are colored differently for these, and their labels are decorated with the conventional "dagger" (†). This character leads or follows the name, based on layout rules for best legibility. Generally speaking, visible leaf nodes always use the leading format, while internal nodes and those "upward" of the current target node use the following format.

The dagger is also shown for an extinct node in the right-hand property inspector, always before the name.

### DIFF
--- a/webapp/static/js/argus/drawtree.js
+++ b/webapp/static/js/argus/drawtree.js
@@ -154,6 +154,9 @@ function createArgus(spec) {
         "pathColor": "#999",
         "strongPathColor": "#000",
         "labelColor": "#000",
+        "extinctNodeColor": "#c96",  // or maybe #963
+        "extinctPathColor": "#c96",
+        "extinctLabelColor": "#a74",
         "provenanceHighlightColor": "#3333ff",
         "provenanceHighlightLabelColor": "#3333ff",
         "provenanceHighlightLabelBackgroundColor": "#ffffff",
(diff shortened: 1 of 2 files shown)


jimallman:
Oh, this is currently deployed and working on **devtree**, [see for example Cetacea](https://devtree.opentreeoflife.org/opentree/argus/ottol@698424/Cetacea) and nearby nodes.

mtholder:
this looks very nice (though it does mean that we really need to fix the lack of _Homo sapiens sapiens_ in OTT, as in the new view it will look like humans are extinct)

jimallman:
Hm, I see that the [Carolina parakeet](https://devtree.opentreeoflife.org/opentree/argus/ottol@461606/Conuropsis-carolinensis) is not flagged as extinct. Maybe someone's holding out hope..? :shrug:

mtholder:
nor is the dodo (https://devtree.opentreeoflife.org/opentree/argus/ottol@455687/Raphus-cucullatus). Mauritius is smaller than the Carolinas, and dodos were bigger than parakeets. So, even less room for hope, there...

jar398:
https://github.com/OpenTreeOfLife/reference-taxonomy/issues/206

For dodo etc, either curate manually (there are many many cases so this is not very practical), get the info via the GBIF API (it's not in the taxonomy dump but is in their database), or get it from PBDB

Issue #1063, opened 2016-09-26
Move tree list to main tab


kcranston:
Proposal to simplify the UI so that it is easier to see the tree list, and easier to see details about trees:

1. Rename the 'Metadata' tab to 'Home' and include the tree list on that page. [Rough mockup](https://opentreeoflife.mybalsamiq.com/projects/curatormockups/New%20study%20homepage)
1. The trees tab contains details about each tree, with a forward / back button and a link to open the popup for the viewer (or maybe an in-page viewer for small trees?) [Mockup](https://opentreeoflife.mybalsamiq.com/projects/curatormockups/New%20tree%20tab)

PR #1223, opened 2020-03-05
Simplified curation tabs

jimallman:
This branch addresses #1063 :
 - consolidates Metadata and Trees tabs into a new Home tab; 
 - moves Metadata viewing/editing to a new popup;
 - shows validation warnings in their best new context; and
 - updates all tab-navigation and browser-history logic to match.

This is [working now on **devtree**](https://devtree.opentreeoflife.org/curator/study/view/ot_1164), please take a look and comment here.

### DIFF
--- a/curator/static/css/default.css
+++ b/curator/static/css/default.css
@@ -914,6 +914,11 @@ tr.after-shims th {
 		margin-left: 4em;
 }
 
+#study-metadata-popup {
+    width: 80%;
+    margin: auto -40%;
+}
+
 #new-taxa-popup {
     width: 80%;
     margin: auto -40%;
--- a/curator/static/css/default.css
+++ b/curator/static/css/default.css
@@ -930,6 +930,13 @@ tr.after-shims th {
     margin-bottom: 10px;
 }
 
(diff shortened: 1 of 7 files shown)


snacktavish:
https://devtree.opentreeoflife.org/curator/study/view/tt_214
It looks like it is handling duplicate DOI's nicely!

jimallman:
Thanks @snacktavish ! This is exactly the test case I needed. This demonstrates the expected behavior:

 -The duplicate-DOI warning appears in both the core-metadata summary _and_ in the detailed editor.
 - It updates in real time (in both places) as the user edits the DOI. As before, this test takes effect when the DOI edit field loses focus.
 - The warning is always followed by a list of links to the "conflicting" studies.

This last point is worth discussion. If desired, we could show these links only in the detailed metadata editor.

jimallman:
These latest changes are [live on **devtree**](https://devtree.opentreeoflife.org/curator/study/view/ot_1164/). The updated suggestions list looks like this (prompt for login/edit appears **if** the user is not already editing):

Screen Shot 2020-03-12 at 11 11 38 AM

PR #1225, opened 2020-04-13
Brute-force changes to block login for feedback.

jimallman:
Pretty much as described. Omits the Login link from the commenting/feedback UI, and removes the underlying URLs just to be make sure they can't log in regardless (e.g. by URL guessing).


jimallman:
Probably time to close this, since it's not an elegant/switchable feature. But probably wise to keep it around Just In Case...

PR #1228, opened 2020-04-21
Generic phylopic err

mtholder:


jimallman:
Since `read-only` is a dead branch, do we need to move this to `master` as well? I'll propose this PR and see what happens...

jimallman:
Since this branch is no longer available, I [applied the 2nd change](https://github.com/OpenTreeOfLife/opentree/commit/05597fad24a468b78bab10efa1291703702c643a) as a fix directly on `master`. The first change (funky import command) was already fixed there.

PR #1249, opened 2020-07-23
SVG inkscape fix

jimallman:
Make the SVG namespace explicit in our internal (borrowed) stylesheet, so that tree styles and colors are recognized by the popular Inkscape program. This is tested and working on **devtree**.


snacktavish:
Looks spiffy! I tried on devtree.

PR #1252, opened 2020-09-24
Test for script-managed trees, show warning msg

jimallman:
As discussed, we should detect studies with script-managed trees and flag them in the curation app. If such a tree is found, this message will appear instead of the normal tree list:
Screen Shot 2020-09-24 at 1 20 07 AM

The test function appears to be working on **devtree**, but this PR is **blocked** since such a study won't load properly in the curation web-app. See for example https://devtree.opentreeoflife.org/curator/study/view/tt_217
The footer message that appears points to a Nexson validation error:
```
Exception in coercing to the required NexSON version for validation.
```

### DIFF
--- a/curator/static/js/study-editor.js
+++ b/curator/static/js/study-editor.js
@@ -9139,6 +9139,19 @@ function currentStudyVersionContributedToLatestSynthesis() {
     return (viewModel.startingCommitSHA === latestSynthesisSHA);
 }
 
+function studyContainsScriptManagedTrees() {
+    // check for signature for this in Nexson (to modify UI, hide/block some features?)
+    var allTrees = viewModel.elementTypes.tree.gatherAll(viewModel.nexml);
+    var bigTrees = ko.utils.arrayFilter(
+        allTrees,
+        function (tree) {
+            // if this property is found (even if it's an empty object), assume it's a huge script-managed tree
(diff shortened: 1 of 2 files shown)


jimallman:
Two possible enhancements come to mind, comments welcome:

- Should we show a tree list if any "normal" trees are also found in the study? So far we've assumed that this will never happen, but we could easily test for both types and show another warning if this is found.

- Should we show script-managed trees in a simplified tree list? This would mainly be useful so that a curator can toggle such trees into/out of the main synthesis collection. (Perhaps this is best done by experts with the raw tree data in hand?)

jimallman:
We now have sensible behavior (in all tabs) for studies with script-managed-trees. Since these have no internal otus, the Analysis tools will be disabled. Working now on **devtree**, see these examples:

- a study with one script-managed tree: https://devtree.opentreeoflife.org/curator/study/edit/tt_217/?tab=analyses

- a study with multiple normal trees (ie, otus stored in Nexson): https://devtree.opentreeoflife.org/curator/study/view/tt_266/?tab=analyses

- a study with no trees at all (behaves like first case above): https://devtree.opentreeoflife.org/curator/study/view/tt_256/?tab=analyses

jimallman:
Stand by on this PR until we have a more satisfying way to handle export options.

PR #1263, opened 2021-05-05
Import TNRS nameset in curation

jimallman:
Here's my initial implementation of nameset merging from the study curation app. This is working now on **devtree**, and ~~I'm currently testing~~ has been tested with a variety of valid and invalid input.

The current behavior is as follows (with _my questions in italics_):

- OTU labels are matched with mapped names based on their "Original name" `ot:originalLabel` field, after trimming leading and trailing whitespace from each and converting them to lower case.
- We currently ignore the "Modified for mapping" label (`adjustedLabel`) of a study OTU. _Is this appropriate?_
- We skip any OTU that has been previously mapped "by hand", as this presumably reflects a more careful mapping.
- To avoid dealing with ambiguous taxon names, we require a mapped name **and** ottid in all formats. _Is this reasonable? If not, should we prompt the curator to resolve ambiguous taxon names with the usual UI? Also, note that even detecting the presence of an ambiguous taxon name would be tricky._
- Despite the current list filters and pagination, we attempt to map all un-mapped OTUs throughout the study. Bulk-mapping results will be reflected in the filtered list.
- After any successful merge operation (at least one OTU successfully mapped), we add a brief message in a shared annotation for merged namesets. These are committed in tandem with the merged names, so it should be fairly easy to trace where things came from. Here's [an example from an "uploaded" ZIP](https://github.com/OpenTreeOfLife/phylesystem-0/commit/d6b60bc590534821ce1671798aa5b649d2c6ef8a), and [another from pasted CSV/text](https://github.com/OpenTreeOfLife/phylesystem-0/commit/688f3c53db152d8ab688792e384fa618fa482812). 
- The current implementation is entirely done with client-side Javascript. If we want a server-side equivalent, I can provide pseudo-code that describes this process exactly.
- _Bonus question: Do we ever expect duplicate OTU labels in a study? (I think not.) If so, we'll need to decide whether to map both or choose one._
- _Do we have any concerns about pollution of data with bogus/mistaken taxon names + ottids? This can easily happen if the user's input (esp. pasted text) is unreliable for some reason._


jimallman:
I've finished testing with variety of pasted and uploaded data. I believe we now return sensible error results or error messages in all cases.

mtholder:
Looks great, Jim. Thanks.
I think the default behaviors that you've used sound good. Some formats (NeXML, for example) do technically allow multiple OTUs to have the same label, but I've never seen analysis software support that. Nor have I seen files "in the wild" that do that. So, I think we can ignore that as a wrinkle to be dealt with.

jimallman:
> We currently ignore the "Modified for mapping" label (adjustedLabel) of a study OTU. Is this appropriate?

I just realized this conflicts with some of the scenarios we discussed on yesterday's call with @LunaSare. We might want to reconsider this, based on actual curator experience. An example might be if a lab uses its own naming conventions, but there's some variation from tree to tree that we can normalize with the regex tools in OTU Mapping.

LunaSare:
Thanks @jimallman! I think adjusted labels do not need to be exposed for now. As long as they are stored somewhere, users can check what label was used for tnrs matching. I think the most common workflow is to go from “original label” to “Ott matched label” and vice versa. I’ll try the new nameset tool today, exciting!!

PR #1264, opened 2021-05-12
Auto-accept exact matches and synonyms

jimallman:
This has been a long-standing request for both the **OTU Mapping** tab (study curation) and the bulk-TNRS tool. It's working now in both web apps on **devtree**. This largely mimics the behavior of the existing **Accept exact matches** button, but also auto-accepts any exact match for a taxon's synonym.


mtholder:
Hi @jimallman, after I merged the other PR, there is now a conflict between this branch and master. It looks like its in a commented out line, but I don't know why they would both be touched and I don't understand the base64 content that is in conflict.
Can you take a look and resolve the conflict?
thanks,
Mark

jimallman:
Thanks for the heads-up! Things should be working fine now.

FYI, the `tnrs-bundle.js` file is produced by npm tools [as described here](https://github.com/OpenTreeOfLife/opentree/blob/master/curator/static/js/tnrs/README.md). The bundle is rebuilt in my local workspace after every change, so it can behave weirdly when changing branches or merging. Sometimes this manifests as a race condition (file has not yet changed after a prior commit), but apparently it can also trigger a merge conflict, if the changes are trivial in source files but obscure in the bundle file.

PR #1265, opened 2021-05-26
Custom curator url for webapp-only installation

kcranston:
For nexttree, we want to install opentree with a tree viewer only, no curation app. This PR (which is all work done by @mtholder ) removes the hard-coded `/curator` and allows for a config-based URL for the curation app, so that the app and docs on nexttree can point to - for example - `tree.opentreeoflife.org/curator`. 

I'd like to get this merged so that we can deploy nexttree from the master branch.


kcranston:
I need to fix the conflicts, and then test the fixes before we can merge this.

jimallman:
@kcranston The conflicts look pretty straightforward, but let me know if you want a second pair of eyes.

PR #1267, opened 2021-06-22
new stats

mtholder:
13.4 is now being served on tree, but I forgot that I'd need to merge and deploy this before the metadata pages load correctly.
Can you merge this if it looks OK to you, @jimallman ?    Thanks,
Mark


jimallman (approved):
Looks good to me! Both the modified files parse as JSON, including the newest records. All figures and dates looks reasonable and properly formatted.

PR #1280, opened 2022-04-19
pinning

mtholder:
I think https://pypi.org/project/jwcrypto/0.9.1/ is the last version (planned) to support python 2.7 v1.0 is pulled down without pinning, and it is definitely not py2.7 compatible


jimallman:
[Related discussion](https://stackoverflow.com/questions/68901591/error-when-importing-jwcrypto-in-python2-7). Thanks for the heads-up!

PR #1281, opened 2022-06-29
Fix for frozen bulk-TNRS w/o trailing slash

jimallman:
Fixes issue https://github.com/OpenTreeOfLife/feedback/issues/548


jimallman:
@snacktavish This will be available for testing on **devtree** the next time we re-deploy webapps.

PR #1282, opened 2022-07-07
data license info

snacktavish:


snacktavish:
A gitter Q made me realize we are not very clear about our license. Does this look OK? I copied the license text from the Hinchliff et al PNAS paper.

jimallman:
This looks good to me. Just two thoughts:

- I generally try to add a link to the CC0 waiver whenever it's mentioned:
   https://creativecommons.org/publicdomain/zero/1.0/

- Can we support the claim that all our data is CC0? I know it's implicit in all submissions, but elsewhere we talk about [the possibility of data with prior licenses](https://github.com/OpenTreeOfLife/opentree/wiki/Submitting-phylogenies-to-Open-Tree-of-Life#adding-a-new-study):
   
> You have the option of importing trees from [TreeBASE](http://treebase.org/) or uploading from your computer. When importing, you need to say something about the licensing for this data. Unless the data has a pre-existing license from another source, we recommend the [CC0 copyright waiver](https://creativecommons.org/choose/zero/), which facilitates re-use and eventual deposition in [Dryad](https://datadryad.org/) for archival. Trees from TreeBASE have no existing license. Tree from Dryad have a CC0 waiver.

jar398:
Legally, cc0 is a rights waiver, not a license (because in the US there are no statutary rights restricting copying of data in the first place, ergo nothing to license). (cc0 has language that makes it revert to a license in jurisdictions that do not permit waiving rights, but in the US this language does not apply.)
So I would not call the section (above the changed line) 'data licenses'.

The text that Jim quotes has a similar problem. To be persnickety it would be (for example) 'the disposition of rights related to this data' rather than 'licensing for this data' and (for example) 'pre-existing license or waiver' instead of 'pre-existing license'.

The changed line seems fine since it talks only about software, to which licensable rights such as copyright do apply.

jimallman:
Thanks for clarifying, @jar398! I've modified the page on **devtree** to incorporate your feedback, see https://devtree.opentreeoflife.org/about/licenses

snacktavish:
Looks great!

PR #1283, opened 2022-08-19
Fix delete collection button

jimallman:
Tiny but important fix for Delete Collection button on single-page collection editor. Please merge to master, per our usual policy for web-appp repo.


jimallman:
This fix is already in place (and tested) on **devtree**

PR #1286, opened 2022-09-01
Disable Treebase import

jimallman:


mtholder:
treebase is now back online, fwiw

jimallman:
Keeping this PR open, in case TreeBASE goes down again.

PR #1292, opened 2023-01-06
fix of misleading error message on misconfigure

mtholder:


mtholder:
This second commit (hardcoding python3 as the executable) was also necessary for me to get the phylogenetic imports working on devtree

Issue #1288, opened 2022-09-30
Curator blocks taxon addition of existing names


snacktavish:
e.g. https://github.com/OpenTreeOfLife/reference-taxonomy/issues/391

but it also seems to store names that you just typed, but didn't submit, and reject those as 'already proposed' as well.
Given that we already have homoyns anyhow, I think a less gated taxon addition process would be good.

jimallman:
@snacktavish, OK to close this? I think it's working as expected on **devtree**.

snacktavish:
Awesome! and Yes. I meant to test on dev tree, but didn't get a chance. Thanks!!

PR #1303, opened 2023-02-28
Allow reusing taxon names

jimallman:
As discussed, we will allow a curator to propose a new taxon with a name that already exists in OTT. (A warning will be displayed, but we won't block submission of the name.) This is tested and working now on **devtree**.


jimallman:
Fixes #1288

Issue #1302, opened 2023-02-14
Should we always mint a new ID when creating a tree or study?


jimallman:
This might avoid confusion in our tree-collection feature, if it can mis-match a new study/tree because serial IDs are being re-used. (E.g., If I remove tree2 from a study, then create a new tree, does it re-use the ID?)

IF this is happening, we should save the last-used IDs for each and ALWAYS increment.

jimallman:
OK, so we [do this properly for new studies](https://github.com/OpenTreeOfLife/peyotl/blob/32a7e0adfa02f8940d266a99d9dbc6e12f566cad/peyotl/phylesystem/phylesystem_shard.py#L321-L330). Each is issued a unique, newly-minted ID, and [a persistent counter is incremented](https://github.com/OpenTreeOfLife/peyotl/blob/32a7e0adfa02f8940d266a99d9dbc6e12f566cad/peyotl/phylesystem/phylesystem_shard.py#L281-L296). So all good there. I'll check tree IDs next.

jimallman:
@snacktavish The feature branch above (currently working on devtree) mints new tree IDs by a completely different method. Instead of simple incrementing IDs (`tree1`, `tree2`...) it tries to make each one unique in the current studio, but also _unlikely_ (`tree478290`) to handle a sequence like 
- add/edit tree4 
- add to a tree collection
- delete tree4 from study
- add new tree4 to study 
-  collection now contains an unintended tree

The only safe alternative I can think of would be for each study to keep track of all IDs it has used or attempted to use previously. I suppose keeping a old-school counter would work, but still fails in the event of a tree that's created, added to a collection, then the curation session is abandoned without saving. So we're back to needing to force a save of the study before adding any of its trees to a collection. Seems weird, but maybe the lesser of evils if we don't like this.

jimallman:
@snacktavish Oh, one minor usability query: Would these six-digit tree IDs like `tree875920` just make it painful to talk about trees? I mean, `tree2` is at least easy to say and find. Maybe there's a still-unlikely compromise, like `tree4839`?

snacktavish:
I'll take a look, but I def think single digits are highly preferable, because there are several places where a human needs to read and track them. Maybe even worth the readability at the risk of tree number re-use.

PR #1304, opened 2023-02-28
Use a simpler scheme for safe element IDs

jimallman:
As discussed, we'd like to have single- or double-digit IDs for elements (esp. trees). This scheme will keep track of the highest minted IDs for each element type and store it in phylesystem. Addresses #1302.


jimallman:
I'm still testing this on **devtree**, so holding it in draft mode for now.

jimallman:
OK, this seems to be working as intended on **devtree**. Studies now save with accurate highest-minted-id information, reload properly, and skip any minted-then-deleted IDs. Done, I think.

jimallman:
See [the history of study tt_266](https://devtree.opentreeoflife.org/curator/study/edit/tt_266/?tab=history) for a series of test commits.

PR #1306, opened 2023-04-04
Conflict with source tree2

jimallman:
Adapt the study+tree browsing UI to support conflict reporting vs. other trees in phylesystem. This seems to be working now (on **devtree**) in all these contexts:

- In the single-tree view popup (conflict details)
- In the Analyses tab for a study (conflict summary)
- In the tree-collection editor, the original context for study+tree lookup (adding tree to a collection)

### DIFF
--- a/curator/static/js/curation-helpers.js
+++ b/curator/static/js/curation-helpers.js
@@ -929,6 +929,7 @@ function getPhylesystemLookupPanel( context ) {
             console.error("getPhylesystemLookupPanel(): UNKNOWN context '"+ context +"'!");
             return null;
     }
+    return $container;
 }
 
 var showingResultsForStudyLookupText = '';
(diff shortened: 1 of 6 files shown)


bredelings:
We should merge this soon!

@snacktavish did find one bug here: #1316

Should we merge now and fix this one afterwards?

jimallman:
I'm chasing down a number of subtle bugs from our introduction of curated reference trees. The tree-display code also generates these URLs and descriptions, and some functions assumed only the original two reference trees (OT taxonomy or the latest synthesis). I'll have this up for review soon.

jimallman:
UPDATE: I've tracked down and fixed one annoying bug: Choosing a new second tree from phylesystem would fail with footer message "Please choose a target (reference) tree for comparison". Now it works as expected.

snacktavish:
I will poke around on dev, and either post a bug here or merge!

jimallman:
One outstanding bug remains (that we can see): If we choose a second tree and the conflict UI reverts to "Choose a target"... instead of keeping its second-tree prompts. This seems to happen whether the 2-tree comparison is successful or not.

snacktavish:
https://github.com/OpenTreeOfLife/opentree/pull/1328
Deployed on dev.

PR #1312, opened 2023-05-02
Expect the curation app base URL in webapp config

jimallman:
This restores the unusual (production) curation app URL used in our "curatorless" synth-tree viewer. Other changes here will reflect any renaming of API "domains" and method URLs (endpoints) happening in ot-ansible.

### DIFF
--- a/curator/controllers/collection.py
+++ b/curator/controllers/collection.py
@@ -14,7 +14,7 @@
 #########################################################################
 
 from applications.opentree.modules.opentreewebapputil import(
-    get_opentree_services_method_urls, 
+    get_opentree_api_endpoints, 
     fetch_current_TNRS_context_names,
     fetch_trees_queued_for_synthesis,
     get_maintenance_info)
@@ -27,7 +27,7 @@ def index():
     Show list searchable/filtered list of all collections
     (default filter = My Collections, if logged in?)
     """
-    view_dict = get_opentree_services_method_urls(request)
+    view_dict = get_opentree_api_endpoints(request)
     view_dict['maintenance_info'] = get_maintenance_info(request)
     if auth.is_logged_in():
         # user is logged in, filter to their own collections by default?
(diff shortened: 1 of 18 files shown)


jimallman:
I've added a number of JS tweaks to match the variables spawned [from our new app config files](https://github.com/OpenTreeOfLife/ot-ansible/pull/32).

snacktavish:
image
It looks like the conflict call in curator on devtree is pointing to api instead of devapi.... which I don't think it should be doing. Is that connected to this PR not being merged? I think that everything in the curator should be pointing to dev, although in the devtree tree viewer, we are pointing to production for the study gets.

snacktavish:
image
I am confused though... because it looks like it should be using the default, which is in this case devapi, for conflict.

jimallman:
Thanks! I can chase this tonight, but if you're curious, check the Network tab (in browser's dev tools) to see where the request actually went.

snacktavish:
Hmmmmm. It is calling devapi.... Request URL is  
https://devapi.opentreeoflife.org/v3/conflict/conflict-status?tree1=pg_2827%23tree6576&tree2=tt_268%23tree1&use_cache=false
but then devapi is calling production phylesystem. :thinking:

bredelings:
Maybe look at `git grep apis_production_base`.  I think this means "use production even on devapi". 

Some of these should probably be `apis_common_base`, which I think means "use dev services on devapi, and productions services on api".  Or something like that.

bredelings:
I think I have a local change somewhere that does this:
```
diff --git a/roles/ws_wrapper/templates/0_config_app.j2 b/roles/ws_wrapper/templates/0_config_app.j2
index 0357d5d..0fadda1 100644
--- a/roles/ws_wrapper/templates/0_config_app.j2
+++ b/roles/ws_wrapper/templates/0_config_app.j2
@@ -14,7 +14,7 @@ pyramid.default_locale_name = en
 # pyramid.includes = pyramid_beaker
 # pyramid.includes = pyramid_debugtoolbar
 
-phylesystem-api.host={{ apis_production_base_URL }}
+phylesystem-api.host={{ apis_common_base_URL }}
 #phylesystem-api.port=
 phylesystem-api.prefix=v3
```

Issue #1321, opened 2023-07-07
"Look up doi" button weirdness


snacktavish:
"Look up doi" button works fine on dev, but just links out to crossref on production. Maybe related to phylesystemapi update somehow?

snacktavish:
Nevermind! I tried again and it worked fine.... ? I will leave open for a bit to see if I can make it happen again and track

jimallman:
_Copying this from Gitter chat:_

Clicking the 'Look Up DOI...' button will fall back to CrossRef if the 'Publication reference' field is empty. Any visible text in this field should do the normal popup dance with matching DOIs.

In other words, this is expected (by me 😸) behavior unless there's text in the 'Publication reference' field. Would you liek to see a different response with there's no ref text, maybe a prompt in the site footer?

snacktavish:
Ahhh! I wanted it to look up the reference text form the doi, and I thought it would do it. I know if there is text in the reference field you can correct it from the doi, but maybe you need to paste the DOI into reference field not the doi field first?

snacktavish:
To further clarify - I had pasted in the DOI when I created the study, but for some reason it didn't autofill any text into the ref. I wanted it to try again on the auto fill - and if I pasted the DOI into the reference text field, it did fill in the whole reference. Can it do that if the DOI is just in the DOI field as well?

jimallman:
@snacktavish The new behavior (submitting DOI field to CrossRef, along with the current reference text) is live now on **devtree**. Please give it a try and, if you like the change, maybe close this and [merge the related PR](https://github.com/OpenTreeOfLife/opentree/pull/1323).

PR #1323, opened 2023-08-08
Smarter DOI lookup

jimallman:
This addresses #1321 by including the current DOI field when searching CrossRef. See related comments there.

### DIFF
--- a/curator/controllers/default.py
+++ b/curator/controllers/default.py
@@ -758,7 +758,7 @@ def to_nexson():
 def search_crossref_proxy():
     encoded_ref_string = request.env.web2py_original_uri.split('search_crossref_proxy?')[1]
     # prepend the CrossRef search base URL and return the response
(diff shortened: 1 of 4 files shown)


snacktavish:
Crossref seems to struggle when looking up just the DOI in URL form, e.g. http://dx.doi.org/10.1111/1365-2656.13565, and it needs to be specifically informed that it is looking up a doi.

jimallman:
Note that on the CrossRef website, a metadata search only [works for a "bare" DOI](https://search.crossref.org/?q=10.1111%2F1365-2656.13565&from_ui=yes), like `10.1111/1365-2656.13565`, but [**fails** for the URL form](https://search.crossref.org/?q=http%3A%2F%2Fdx.doi.org%2F10.1111%2F1365-2656.13565&from_ui=yes) we generally prefer.

Issue #1299, opened 2023-02-10
Collections "save a copy" doesn't work


snacktavish:
Pressing the "save a copy" button on a collection doesn't seem to do anything.

snacktavish:
Re-upping this one!

jimallman:
This should be working as expected on **devtree**, including opening the _new_ copy (instead of the original) in the collection editor.

snacktavish:
Huh, have we maybe not merged or maybe not deployed since this was fixed?

jimallman:
Here's (I believe) the related pull request, merged to master back in March: https://github.com/OpenTreeOfLife/opentree/pull/1336

PR #1336, opened 2024-02-20
Fix collection Save Copy

jimallman:
When saving a copy of a tree collection, it seems sensible that we should jump to the editor of the *new* (saved) collection, instead of staying put. This should now be working (with some other bug fixes) in the full-page *and* popup editor. Working now on **devtree**.


snacktavish:
It complains that I am missing a public Github email, but only on save a copy of collection.

jimallman:
Addresses #1299

